PR #4, opened 2019-08-07
 Set published source for sync-db

shradayshakya:
`yarn add @leapfrogtechnology/sync-db`


examples/node-app-mssql/package.json
```diff
@@ -13,6 +13,7 @@
   },
   "license": "MIT",
   "devDependencies": {
+    "@leapfrogtechnology/sync-db": "^1.0.0-alpha.1",
```
mesaugat: This should be in `dependencies`.

PR #6, opened 2019-08-09
Remove password attribute from connections object before logging

SafalPandey:
* Remove password attribute from connections object before logging


src/util/object.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Get the copy of object without attributes.
+ *
+ * @param {any} obj
+ * @param {any[]} attrsToExclude
+ * @returns {T}
+ */
+export function withoutAttrs<T>(obj: any, attrsToExclude: any[]): T {
```
mesaugat: This is an overkill at the moment.

mesaugat: Let's remove this util.

src/config.ts
```diff
@@ -52,7 +53,7 @@ export async function resolveConnections(): Promise<Connection[]> {
     id: connection.id || `${connection.host}/${connection.database}`
   }));
 
-  log('Resolved connections: %O', connections);
+  log('Resolved connections: %O', listWithoutAttrs<Connection>(connections, ['password']));
```
mesaugat: Let's map through the connections and only output the database name.

src/config.ts
```diff
@@ -43,7 +44,7 @@ export async function resolveConnections(): Promise<Connection[]> {
   const loaded = await fs.read(filename);
   const { connections } = JSON.parse(loaded);
 
-  log('Connections parsed: %o', connections);
+  log('Connections parsed: %o', listWithoutAttrs<Connection>(connections, ['password']));
```
mesaugat: No need to log anything here.

src/util/object.ts
mesaugat: Also, if we are to add such utils we will need tests for it.

src/config.ts
```diff
@@ -43,16 +43,21 @@ export async function resolveConnections(): Promise<Connection[]> {
...
-
   // TODO: Validate the connections received from file.
 
   const result = connections.map((connection: Connection) => ({
     ...connection,
     id: connection.id || `${connection.host}/${connection.database}`
   }));
 
-  log('Resolved connections: %O', connections);
+  log('Resolved connections: %O', connections.map((connection: Connection) => {
```
kabirbaidhya:
```ts
({ id, host, database }) => ({ id, host, database })
```

PR #23, opened 2019-10-26
Setup docker and docker-compose configuration for the node-app-mssql example

### LINKED ISSUE #8: Setup docker and docker-compose configuration for the node-app-mssql example
Setup docker and docker-compose configuration for the node example demonstrated here - https://github.com/leapfrogtechnology/sync-db/tree/master/examples/node-app-mssql. 

The example projects should be simply runnable just a simple command. 
```bash
$ cd examples/node-app-mssql
$ docker-compose up
```

silwalanish:
Resolves #8 
- Add `Dockerfile`
- Add `docker-compose.yml`
- Add `docker-start` script to run on docker

**Output**
image


examples/node-app-mssql/README.md
```diff
@@ -50,3 +50,14 @@ Calculations:
...
+
+## Docker
+Set `DB_PASSWORD` (password for `SA` user) in environment. e.g 
+```bash
+$ export DB_PASSWORD=Password@123
+```
+Then
+```bash
+$ docker-compose up mssql
+$ docker-compose up app
```
kabirbaidhya: You're already defined a dependency for `mssql` in the `docker-compose.yml` file so you can now just do this:
```
$ docker-compose up app
```

kabirbaidhya:
@silwalanish Awesome. Thanks :1st_place_medal: . 

Can you also post the screenshot of how it looks in the PR description? Thanks :)

examples/node-app-mssql/README.md
silwalanish: We need to run `$ docker-compose up app` only after `mssql` has started. Thus keeping this as it is.

And I removed `mssql` from `depends_on` because it would start `mssql` again even if it was running. And that caused `sync-db` script to not connect to db.

examples/node-app-mssql/connections.sync-db.json.docker
```diff
@@ -0,0 +1,13 @@
+{
+  "connections": [
+    {
+      "id": "db1",
+      "host": "mssql",
+      "port": 1433,
+      "user": "sa",
+      "database": "master",
+      "password": "password",
```
kabirbaidhya: Use the same password as you've shown in `README` so that the user can run it w/o making any changes by default. 
```suggestion
      "password": "Password@123",
```

examples/node-app-mssql/README.md
kabirbaidhya: Makes sense.

examples/node-app-mssql/package.json
```diff
@@ -5,7 +5,8 @@
     "lint": "eslint src/index.js",
     "lint:fix": "eslint --fix src/index.js",
     "sync-db": "sync-db",
-    "start": "yarn lint && node src/index.js"
+    "start": "yarn lint && node src/index.js",
+    "docker-start": "yarn sync-db && yarn start"
```
kabirbaidhya: No need this here. I think you can put this directly in the `Dockerfile` so that the users know the commands explicitly.

examples/node-app-mssql/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:10
+
+WORKDIR /app/
+
+COPY . .
+
+RUN yarn
+
+CMD ["yarn", "docker-start"]
```
kabirbaidhya: Explicitly adding `yarn sync-db && yarn start` would make for sense for the users so that they know first the `sync-db` cli is run to synchronize the database and only then the app is started.

silwalanish: Done

examples/node-app-mssql/package.json
silwalanish: Done

examples/node-app-mssql/connections.sync-db.json.docker
silwalanish: Done

PR #28, opened 2019-11-28
Throw an error if object type names doesn't match directory convention

### LINKED ISSUE #5: Logs are misleading in rollback / teardown stage
## Issue

When the folder names where the database objects are in do not match one of `procedure`, `schema`, `view` or `function`, the logs explicitly say `Rolling back <object name>` and `Executed: <xx> ms` However the objects aren't removed from the database.

## Expected behavior

I think it should raise an error instead. We need to make it clear that the directory name did not match the expected convention and raise an error if rollback script is not executed.

## Log trail

```
  sync-db:db1 Synchronize database +0ms
  sync-db:db1 Running rollback on a database: db1 +0ms
  sync-db:db1 Rolling back: etl.noop +0ms
  sync-db:db1 Executed:   +48ms
  sync-db:db1 Rolling back: etl +0ms
  sync-db:db1 Executed:   +2ms
  sync-db:db1 Finished running rollback +273ms
  sync-db:db1 Running setup +0ms
  sync-db:db1 Running schemas/etl.sql +0ms
  sync-db:db1 Running procedures/etl/noop.sql +0ms
  sync-db Error caught:  { TransactionError: Transaction has been aborted.
...
```

p0k8h:
Resolves #5

- Added `if` block which throws an error message if `dropStatement` is empty.
- Added test case to cover it.


src/util/ErrorHandler.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Custom ObjectTypeError which extends Error class.
+ */
+class ObjectTypeError extends Error {
```
mesaugat: Why is this named `ErrorHandler.ts` when the class name is `ObjectTypeError`? Also, Can we not simply do `throw new Error()`?

src/sqlRunner.ts
```diff
@@ -94,7 +95,10 @@ export function extractSqlFileInfo(filePath: string): SqlFileInfo {
 export function getDropStatement(type: string, fqon: string): string {
   const dropStatement = dropStatementsMap[type];
 
-  return dropStatement ? `${dropStatement} ${fqon}` : '';
+  if (dropStatement) { return `${dropStatement} ${fqon}`; }
```
mesaugat: I'd like this to be the other way round so that it is clear what the function is returning.

```js
if (!dropStatement) {
  const message = `Naming convention must be exact as the directory names inside src/`;
  
  throw new Error('No database object type: ${type}. ${message}');
}

return `${dropStatement} ${fqon}`;
```

src/sqlRunner.ts
```diff
@@ -8,6 +8,7 @@ import Mapping from './domain/Mapping';
 import SqlCode from './domain/SqlCode';
 import * as promise from './util/promise';
 import SqlFileInfo from './domain/SqlFileInfo';
+import ObjectTypeError from './util/ErrorHandler';
```
kabirbaidhya: .

src/util/ErrorHandler.ts
kabirbaidhya: This file isn't needed.

p0k8h: To give specific error type i.e: `ObjectTypeError` i created the file

Also, I feel this might not be needed.

Issue #31, opened 2019-12-06
New CLI arg to generate connections file using the environment variables


kabirbaidhya:
**CLI Usage**
```
$ sync-db --generate-connections
```
**Requirements**
1. Introduce a CLI argument / flag `--generate-connections` that if invoked with - should generate a `connections.sync-db.json` file.
2. The file should be generated in the path same as `sync-db.yml`.
3. If successful it should print out a message with the generated file path.
4. In case of failure it should display the error message of why it failed.
5. Use these environment variables to get the credentials:
   - `DB_HOST` - Database Host
   - `DB_PASSWORD` - Password
   - `DB_NAME` - Database name
   - `DB_USERNAME` - Database username
   - `DB_PORT` - Port
   - `DB_CLIENT` - Database client eg: `mssql`, `mysql`
   - `DB_ENCRYPTION` - Boolean value to set whether or not to use encrypted connection. Should be `false` by default unless explicitly set as `true`.

---
**Proof of concept / demo** - the feature explained above should work in the same way as this demo bash script below.
```bash
echo '{
  "connections": [
    {
      "client": "'$DB_CLIENT'",
      "host": "'"$DB_HOST"'",
      "port": '"$DB_PORT"',
      "user": "'"$DB_USERNAME"'",
      "password": "'"$DB_PASSWORD"'",
      "database": "'"$DB_NAME"'",
      "options": {
        "encrypt": '"${DB_ENCRYPTION:=false}"'
      }
    }
  ]
}' > 'connections.sync-db.json'
```

p0k8h:
I'll be working in it

PR #33, opened 2019-12-13
CLI arg to generate connections from the environment

p0k8h:
- Added `--generate-connections` flag to generate `connections.sync-db.json` file reading data from ENV.
- Added syncDb service which generates connections by reading values from ENV.
- Added fs util method to write file.
- Added interfaces for syncDbOptions(flags) and OclifLogger.
- Added test suite with cases for fs utility module.
------

Command :
```bash
$ sync-db --generate-connections
```
Resolves https://github.com/leapfrogtechnology/sync-db/issues/31


README.md
```diff
@@ -56,9 +56,10 @@ USAGE
   $ sync-db
 
 OPTIONS
-  -f, --force    Force synchronization
-  -h, --help     Print help information
-  -v, --version  Print version
+  -f, --force            Force synchronization
+  -h, --help             Print help information
+  -v, --version          Print version
+  --generate-connections Generate connections
```
mesaugat: Can we have a short flag `-c` to do the same?

src/SyncDb.ts
```diff
@@ -2,6 +2,7 @@ import { Command, flags } from '@oclif/command';
 
 import { log } from './logger';
 import { loadConfig, resolveConnections } from './config';
+import { syncDbServices } from './util/syncDbServices';
```
mesaugat: Ladder. ↕️

src/SyncDb.ts
```diff
@@ -15,7 +16,8 @@ class SyncDb extends Command {
   static flags = {
     version: flags.version({ char: 'v', description: 'Print version', name: 'sync-db' }),
     help: flags.help({ char: 'h', description: 'Print help information' }),
-    force: flags.boolean({ char: 'f', description: 'Force synchronization' })
+    force: flags.boolean({ char: 'f', description: 'Force synchronization' }),
+    'generate-connections': flags.boolean({ description: 'Generate connections' })
```
mesaugat: Add `char: 'c'`

src/SyncDb.ts
```diff
@@ -30,6 +32,11 @@ class SyncDb extends Command {
     };
 
     try {
+
```
mesaugat: Remove new line here.

src/util/fs.ts
```diff
@@ -47,3 +47,21 @@ export function glob(pathName: string): Promise<string[]> {
...
+ * @param {any} data
+ * @returns {Promise<string>}
+ */
+export function write(filepath: string, data: any): Promise<string> {
+  return new Promise((resolve, reject) => {
+    fs.writeFile(filepath, data, err => {
+      if (err) {
+        return reject(err);
+      }
+      resolve();
```
mesaugat: Always return.

```js
return resolve();
```

src/util/generateConnection.ts
```diff
@@ -0,0 +1,67 @@
...
+  NAME: 'database'
+};
+
+/**
+ * Reads values from ENV from provide keys.
+ *
+ * @param {string[]} keys
+ * @returns {string[] | Ikeys}
+ */
+function extractValuesFromENV(keys: string[]): string[] | Ikeys {
```
kabirbaidhya:
```suggestion
function extractFromEnv(keys: string[]): string[] | Ikeys {
```

src/util/fs.ts
```diff
@@ -47,3 +47,22 @@ export function glob(pathName: string): Promise<string[]> {
...
 }
+
+/**
+ * Write contents to file.
+ *
+ * @param {string} filepath
+ * @param {any} data
+ * @returns {Promise<string>}
+ */
+export function write(filepath: string, data: any): Promise<string> {
```
kabirbaidhya:
```suggestion
export function write(filepath: string, data: any): Promise<void> {
```

kabirbaidhya: We need test for this function too.

src/util/types.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface Ilogger {
+  info: (message: string) => void;
+  error: (input: string | Error, options: {
+    code?: string | undefined;
+    exit: false;
+  }) => void;
+}
+
+export interface Imappings {
```
kabirbaidhya:
```suggestion
export interface Mapping<T> {
```

src/util/types.ts
```diff
@@ -0,0 +1,24 @@
...
+  [key: string]: string;
```
kabirbaidhya:
```suggestion
  [key: string]: T;
```

src/util/types.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+export interface Iflags {
+  version: void;
+  help: void;
+  force: boolean;
+  'generate-connections': boolean;
+}
+
+export interface Ilogger {
```
kabirbaidhya:
```suggestion
export interface Logger {
```

README.md
```diff
@@ -56,9 +56,10 @@ USAGE
   $ sync-db
 
 OPTIONS
-  -f, --force    Force synchronization
-  -h, --help     Print help information
-  -v, --version  Print version
+  -f, --force                 Force synchronization
+  -h, --help                  Print help information
+  -v, --version               Print version
+  -c, --generate-connections  Generate connections
```
kabirbaidhya:
```suggestion
  -c, --generate-connections  Generate connections.sync-db.json file.
```

src/util/types.ts
```diff
@@ -0,0 +1,24 @@
...
+    [key: string]: boolean
+  };
+}
+
+export interface Iflags {
+  version: void;
+  help: void;
+  force: boolean;
+  'generate-connections': boolean;
+}
```
kabirbaidhya: I'm not sure if we need this type. If we do, it could be named `CLIFlags` or `SyncDbOptions`.

src/util/fs.ts
p0k8h: Shall I write the test cases in this PR itself?

kabirbaidhya: Yes.

src/util/generateConnection.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+  });
+
+  if (errors.length) {
+    throw new Error(JSON.stringify({ 'keys not found': errors }));
+  }
+
+  return connection;
+}
```
kabirbaidhya: This implementation doesn't look neat enough. Look into some of the existing code here to get some idea:
- https://github.com/leapfrogtechnology/sync-db/blob/master/src/config.ts
- https://github.com/leapfrogtechnology/sync-db/blob/master/src/constants.ts

Some of the things you can reuse already. Also, make this function `extractFromEnv` return `Promise<ConnectionConfig[]>` i.e same as what [resolveConnections](https://github.com/leapfrogtechnology/sync-db/blob/master/src/config.ts#L37) returns.

Also, you can move this function to `config.ts` and name it `resolveConnectionsFromEnv` - because it's similar.

kabirbaidhya: Make it straightforward like the example code here in the issue https://github.com/leapfrogtechnology/sync-db/issues/31.

You can do:
```ts
const connection = {
  client: process.env.DB_CLIENT,
  host: process.env.DB_HOST,
  ...
  options: {
    encrypt: ...
  }
};
```

src/util/fs.ts
```diff
@@ -47,3 +47,22 @@ export function glob(pathName: string): Promise<string[]> {
...
 }
+
+/**
+ * Write contents to file.
+ *
+ * @param {string} filepath
+ * @param {any} data
+ * @returns {Promise<string>}
+ */
+export function write(filepath: string, data: any): Promise<void> {
```
kabirbaidhya: Need tests.

src/util/generateConnection.ts
```diff
@@ -0,0 +1,67 @@
+import * as fs from './fs';
+import { Key, Logger, Mapping } from './types';
+
+const CONNECTIONS_FILE_NAME = 'connections.sync-db.json';
```
kabirbaidhya: Use the esisting constant value.

src/util/generateConnection.ts
```diff
@@ -0,0 +1,67 @@
...
+const KEYS = ['DB_HOST', 'DB_PASSWORD', 'DB_NAME', 'DB_USERNAME', 'DB_PORT', 'DB_CLIENT', 'DB_ENCRYPTION'];
+const mapping: Mapping<string> = {
+  USERNAME: 'name',
+  NAME: 'database'
+};
```
kabirbaidhya: No need.

src/util/generateConnection.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Reads values from ENV from provide keys.
+ *
+ * @param {string[]} keys
+ * @returns {string[] | Key}
+ */
+function extractFromEnv(keys: string[]): string[] | Key {
```
kabirbaidhya: Should return a list of database connections (only has one item for now).
```suggestion
function extractFromEnv(keys: string[]): ConnectionConfig[] {
```

src/util/syncDbServices.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @param {Logger} logger
+ * @returns {Promise<void>}
+ */
+export async function syncDbServices(flags: SyncDbOptions, logger: Logger): Promise<void> {
+
+  if (flags['generate-connections']) {
+    await generateConnection(logger);
+    process.exit(0);
+  }
+}
```
kabirbaidhya: This is not a util.

src/util/generateConnection.ts
p0k8h: Instead of
~~validateConnections(process.env)~~, `validateConnections(conn_keys)`. And access the `process.env` inside the function.

And Loop is compulsory to validate the connections? Will this implementation be Okay?

src/config.ts
```diff
@@ -54,3 +54,46 @@ export async function resolveConnections(): Promise<ConnectionConfig[]> {
...
+
+  const connection = {
+    client: process.env.DB_CLIENT,
+    host: process.env.DB_HOST,
+    port: process.env.DB_PORT,
+    user: process.env.DB_USERNAME,
+    password: process.env.DB_PASSWORD,
+    database: process.env.DB_NAME,
+    options: {
+      encrypt: process.env.DB_ENCRYPTION || false
```
kabirbaidhya:
```suggestion
      encrypt: process.env.DB_ENCRYPTION === 'true' || false
```

src/config.ts
```diff
@@ -54,3 +54,46 @@ export async function resolveConnections(): Promise<ConnectionConfig[]> {
...
+ * Resolve database connections from Env.
+ *
+ * @param {string[]} keys
+ * @returns {ConnectionConfig[]}
+ */
+export function resolveConnectionsFromEnv(): ConnectionConfig[] {
+
+  validateConnections(ENV_KEYS);
+
+  const connection = {
```
kabirbaidhya:
```suggestion
  const connection: ConnectionConfig = {
```

src/config.ts
```diff
@@ -54,3 +54,46 @@ export async function resolveConnections(): Promise<ConnectionConfig[]> {
...
+    port: process.env.DB_PORT,
+    user: process.env.DB_USERNAME,
+    password: process.env.DB_PASSWORD,
+    database: process.env.DB_NAME,
+    options: {
+      encrypt: process.env.DB_ENCRYPTION || false
+    }
+  };
+
+  return [connection as ConnectionConfig];
```
kabirbaidhya: Make the above change and you can make it simple.
```suggestion
  return [connection];
```

src/util/fs.ts
```diff
@@ -47,3 +47,39 @@ export function glob(pathName: string): Promise<string[]> {
...
+ * @param { string } filepath
+ */
+export function remove(filepath: string): Promise<string> {
+  return new Promise((resolve, reject) => {
+    fs.unlink(filepath, err => {
+      if (err) {
+        return reject(err);
+      }
+
+      return resolve(`${filepath} was removed`);
```
kabirbaidhya: No need to return a string message on success. Just a `boolean` value `true` or even `void` would do.
```suggestion
      return resolve();
```

.gitignore
```diff
@@ -7,3 +7,5 @@
 /tmp
 node_modules
 /coverage
+/test/hello.txt
+/fake_file.txt/random
```
kabirbaidhya: Create a directory and ignore it instead of ignoring individual files - could be a `.tmp/` or `test/.tmp` directory.

src/config.ts
p0k8h: Typescript is throwing error

 ```js
Type 'string | undefined' is not assignable to type 'string'.

Type 'undefined' is not assignable to type 'string'.ts(2322)
The expected type comes from property ${KEY} which is declared 
here on type 'ConnectionConfig'
```

```js    
const connection: ConnectionConfig = {}
```
 as `process.env[KEY]` might be `undefined` also. But the key have only `scalar` type in `ConnectionConfig` (No undefined type).

##### As we have already handled the `undefined` case in the above _`validateConnections`_ function.

Using  `{} as ConnectionConfig` shouldn't be a problem here.

```js
const connection = {} as ConnectionConfig;
```
@mesaugat @kabirbaidhya

src/config.ts
```diff
@@ -54,3 +54,48 @@ export async function resolveConnections(): Promise<ConnectionConfig[]> {
...
+    client: process.env.DB_CLIENT,
+    host: process.env.DB_HOST,
+    port: process.env.DB_PORT,
+    user: process.env.DB_USERNAME,
+    password: process.env.DB_PASSWORD,
+    database: process.env.DB_NAME,
+    options: {
+      encrypt: process.env.DB_ENCRYPTION === 'true'
+    }
+  } as ConnectionConfig;
```
p0k8h: https://github.com/leapfrogtechnology/sync-db/pull/33/files#r358632789

src/config.ts
```diff
@@ -54,3 +54,48 @@ export async function resolveConnections(): Promise<ConnectionConfig[]> {
...
+ * @returns {void}
+ */
+function validateConnections(keys: string[]): void {
+  const errors: string[] = [];
+
+  keys.forEach(element => {
+    if (!process.env[element]) {
+      errors.push(element);
+    }
+  });
```
kabirbaidhya: Could be a oneliner.
```ts
const missingVars = keys.filter(key => !process.env[key]);
```

src/config.ts
```diff
@@ -54,3 +54,48 @@ export async function resolveConnections(): Promise<ConnectionConfig[]> {
...
+
+  if (errors.length) {
+    throw new Error(`Following environment variables were not set: ${errors}`);
```
kabirbaidhya:
```suggestion
    throw new Error('Following environment variables were not set: ' + missingVars.join(', '));
```

src/config.ts
kabirbaidhya: This is a reasonable typecheck so we shouldn't suppress this, but instead we should resolve the issue somewhere in the assignment.

kabirbaidhya: You're getting these issue because each of the keys with `process.env.*` can be either `string` or `undefined`. So, you'll need to null-coalesce here.

src/services/syncDbServices.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Services to run depending upon the flags provided.
+ *
+ * @param {SyncDbOptions} flags
+ * @param {Logger} logger
+ * @returns {Promise<void>}
+ */
+export async function syncDbServices(flags: SyncDbOptions, logger: Logger): Promise<void> {
+
```
kabirbaidhya: .

test/fs.test.ts
```diff
@@ -0,0 +1,65 @@
+import 'mocha';
+import * as path from 'path';
+import { expect } from 'chai';
+
+import * as fs from '../src/util/fs';
+
+describe('UTIL: fs', () => {
+
```
kabirbaidhya: .

src/config.ts
```diff
@@ -54,3 +54,48 @@ export async function resolveConnections(): Promise<ConnectionConfig[]> {
...
+}
+
+/**
+ * Resolve database connections from Env.
+ *
+ * @param {string[]} keys
+ * @returns {ConnectionConfig[]}
+ */
+export function resolveConnectionsFromEnv(): ConnectionConfig[] {
+
```
kabirbaidhya: .

test/fs.test.ts
kabirbaidhya: Hint: you can create a physical file (temporary) under our temp directory and cleanup for each of these tests - you can use before / after hooks for that.

cc @mesaugat @SafalPandey

kabirbaidhya: FYI: https://nodejs.org/api/fs.html#fs_fs_mkdtemp_prefix_options_callback

test/fs.test.ts
```diff
@@ -0,0 +1,65 @@
...
+  it('should write content to the filepath', done => {
+    fs.write(filePath, fileContent).then(result => {
+      expect(result).to.equal(undefined);
+      done();
+    });
+  });
+
+  it('should read the content from the filepath', done => {
+    fs.read(filePath).then(res => {
```
kabirbaidhya: How do you know this file exists? This is not a deterministic approach. All tests need to be independent and if it requires something you should do and undo it in setup / teardown phases.

src/domain/OcliffLogger.ts
```diff
@@ -0,0 +1,7 @@
+export default interface OcliffLogger {
```
kabirbaidhya:
```suggestion
interface OclifLogger {
```

src/domain/OcliffLogger.ts
```diff
@@ -0,0 +1,7 @@
...
+  info: (message: string) => void;
+  error: (input: string | Error, options: {
+    code?: string | undefined;
+    exit: false;
+  }) => void;
+}
```
kabirbaidhya: Export default at the bottom.
```ts
export default OclifLogger;
```

src/domain/SyncDbOptions.ts
```diff
@@ -0,0 +1,6 @@
+export default interface SyncDbOptions {
```
kabirbaidhya: same here
```suggestion
interface SyncDbOptions {
```

src/services/syncDb.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @param {OcliffLogger} logger
+ * @returns {Promise<void>}
+ */
+async function generateConnection(logger: OcliffLogger): Promise<void> {
+  try {
+    const filePath = path.resolve(process.cwd(), CONNECTIONS_FILENAME);
+
+    const connections = resolveConnectionsFromEnv();
+    await fs.write(filePath, JSON.stringify({ connections })).then(() => {
```
kabirbaidhya: No need to use `await` and `then` together.
```suggestion
    const contents = JSON.stringify({ connections });
    await fs.write(filePath, contents);
    
    logger.info(`Generated file: ${CONNECTIONS_FILENAME}`);
```

src/services/syncDb.ts
```diff
@@ -0,0 +1,43 @@
...
+      logger.info(`Generated file: ${CONNECTIONS_FILENAME}`);
+    });
+
+    return Promise.resolve();
```
kabirbaidhya: No need.

src/services/syncDb.ts
```diff
@@ -0,0 +1,43 @@
+import * as path from 'path';
+
+import * as fs from '../util/fs';
+import OcliffLogger from '../domain/OcliffLogger';
```
kabirbaidhya:
```suggestion
import Ocliffogger from '../domain/OclifLogger';
```

src/util/fs.ts
```diff
@@ -47,3 +47,40 @@ export function glob(pathName: string): Promise<string[]> {
...
+
+      return resolve();
+    });
+  });
+}
+
+/**
+ * Remove the file.
+ *
+ * @param { string } filepath
```
kabirbaidhya:
```suggestion
 * @param {string} filepath
```

kabirbaidhya:
@p0k8h Pull latest master and resolve conflicts.

src/SyncDb.ts
```diff
@@ -30,6 +32,10 @@ class SyncDb extends Command {
     };
 
     try {
+      await flagConfigs(parsedFlags, {
```
kabirbaidhya: handleFlags

test/fs.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+      filePath = (`${dirPath}/hello.txt`);
+      invalidFilePath = (`${dirPath}/file.txt/random`);
+      await write(filePath, fileContent).catch(e => done(e));
+      done();
+    });
+  });
+
+  it('should write content to the filepath', done => {
+    write(filePath, fileContent).then(result => {
```
kabirbaidhya: await

test/fs.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      read(filePath).then(res => {
```
kabirbaidhya: const res = await read()

test/fs.test.ts
```diff
@@ -0,0 +1,67 @@
...
+        expect(res).to.equal(fileContent);
+        done();
+      });
+    });
+  });
+
+  it('should read the content from the filepath', done => {
+    read(filePath).then(res => {
```
kabirbaidhya: await

test/fs.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      expect(res).to.equal(fileContent);
+      done();
+    });
+  });
+
+  it('should remove the file', done => {
+    remove(filePath).then(() => {
```
kabirbaidhya: await

PR #36, opened 2020-01-19
Config Injection - ability to inject dynamic configurations such that the running SQL transactions could access them

kabirbaidhya:
## Introduce config injection into SQL scripts
- Ability to inject dynamic configurations into the scripts such that the scripts being run via sync-db can access them during the execution. 
- The lifetime of these configurations is limited to only when sync-db is running. And only the scripts directly triggered by `sync-db` can access them. Once they've finished running, the configurations are cleaned up.
- Tested and updated the sample project here :point_right:  https://github.com/leapfrogtechnology/sync-db/pull/37

### How it works?

Add configurations to be injected in the `sync-db.yml` file. It supports environment variable expansion too, so any confidential values / secrets could come through env vars as well.
```yml
hooks:
  post_sync:
    - scripts/setup.sql
injectedConfig:
  vars:
    create_admin: ${CREATE_ADMIN}
    admin_email: 'admin@example.com'
    task_status: ${TASK_STATUS}
```

Now, when `sync-db` is run it creates a temporary table named `__injected_config` which will contain these injected configurations. 

The table would look like this in the run-time. 

| key | value |
|------|---------|
| create_admin | true |
| admin_email | admin@example.com |
| task_status | somevalue |

Now, when the scripts like this runs through `sync-db`, it can access the values in the run-time.

```sql
-- File: ./scripts/setup.sql

--
-- Setup data.
--
DECLARE @create_admin BIT;
DECLARE @admin_email VARCHAR(100);
DECLARE @task_status VARCHAR(100);

SET @create_admin = (SELECT IIF([value] = 'true', 1, 0) FROM __injected_config WHERE [key] = 'create_admin');
SET @admin_email = (SELECT [value] FROM __injected_config WHERE [key] = 'admin_email');
SET @task_status = (SELECT [value] FROM __injected_config WHERE [key] = 'task_status');

-- Setup data.
EXEC testdata.setup_data @create_admin, @admin_email, @task_status;
```

**Note: This table is temporary. Once all the scripts in the transaction have finished running, it will be dropped automatically.**

### DIFF
--- a/src/Connection.ts
+++ b/src/Connection.ts
@@ -82,6 +82,16 @@ class Connection {
   }
 
   /**
+   * Gets access to the schema builder.
+   *
+   * @returns {Knex.SchemaBuilder}
+   * @memberof Connection
+   */
+  public schema(): Knex.SchemaBuilder {
+    return this.instance.schema;
+  }
+
+  /**
    * Returns connection config.
    *
    * @returns {ConnectionConfig}
--- a/src/Connection.ts
+++ b/src/Connection.ts
@@ -94,7 +94,6 @@ class Connection {
    * Gets access to the schema builder.
    *
    * @returns {Knex.SchemaBuilder}
-   * @memberof Connection
    */
   public schema(): Knex.SchemaBuilder {
     return this.instance.schema;
--- a/src/Connection.ts
+++ b/src/Connection.ts
@@ -82,6 +82,15 @@ class Connection {
   }
 
   /**
+   * Returns the underlying Knex instance.
+   *
+   * @returns {Knex}
+   */
+  public getInstance(): Knex {
+    return this.instance;
+  }
+
+  /**
    * Gets access to the schema builder.
    *
    * @returns {Knex.SchemaBuilder}
--- a/src/config.ts
+++ b/src/config.ts
@@ -7,8 +7,8 @@ import * as fs from './util/fs';
 import DbConfig from './domain/DbConfig';
 import SyncConfig from './domain/SyncConfig';
 import ConnectionConfig from './domain/ConnectionConfig';
-import { DEFAULT_CONFIG, CONFIG_FILENAME, CONNECTIONS_FILENAME, ENV_KEYS } from './constants';
 import { prepareInjectionConfigVars } from './services/configInjection';
+import { DEFAULT_CONFIG, CONFIG_FILENAME, CONNECTIONS_FILENAME, ENV_KEYS } from './constants';
 
 /**
  * Load config yaml file.
@@ -25,7 +25,7 @@ export async function loadConfig(): Promise<SyncConfig> {
 
   const loaded = mergeDeepRight(DEFAULT_CONFIG, loadedConfig) as SyncConfig;
 
-  // TODO: Validate the loaded config.
+  validate(loaded);
 
   return {
     ...loaded,
@@ -34,6 +34,17 @@ export async function loadConfig(): Promise<SyncConfig> {
       vars: prepareInjectionConfigVars(loaded.injectedConfig.vars)
     }
   };
+}
+
+/**
+ * Validate the loaded configuration.
+ * TODO: IMPLEMENT THIS.
+ *
+ * @param {SyncConfig} config
+ */
+function validate(config: SyncConfig) {
+  // TODO: Validate the loaded config.
+  // Throw error if validation fails.
 }
 
 /**
--- a/src/config.ts
+++ b/src/config.ts
@@ -51,7 +51,7 @@ function validate(config: SyncConfig) {
   }
 
   // Shouldn't reach under here unless the user has mismatched the value.
-  if (!injectedConfig.vars || isObject(injectedConfig.vars)) {
+  if (!injectedConfig.vars || !isObject(injectedConfig.vars)) {
     throw new Error('Invalid configuration value for `injectedConfig.vars`.');
   }
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,11 +4,11 @@ import { mergeDeepRight } from 'ramda';
 
 import { log } from './logger';
 import * as fs from './util/fs';
-import Mapping from './domain/Mapping';
 import DbConfig from './domain/DbConfig';
 import SyncConfig from './domain/SyncConfig';
 import ConnectionConfig from './domain/ConnectionConfig';
 import { DEFAULT_CONFIG, CONFIG_FILENAME, CONNECTIONS_FILENAME, ENV_KEYS } from './constants';
+import { prepareInjectionConfigVars } from './services/configInjection';
 
 /**
  * Load config yaml file.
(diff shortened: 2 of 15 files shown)


src/Connection.ts
```diff
@@ -81,6 +81,25 @@ class Connection {
...
+   */
+  public getInstance(): Knex {
+    return this.instance;
+  }
+
+  /**
+   * Gets access to the schema builder.
+   *
+   * @returns {Knex.SchemaBuilder}
+   * @memberof Connection
```
mesaugat: Remove `@memberof` ?

src/services/configInjection.ts
```diff
@@ -0,0 +1,109 @@
...
+ */
+export async function setup(connection: Connection, config: SyncConfig): Promise<void> {
+  const logDb = dbLogger(connection);
+  const { injectedConfig } = config;
+
+  logDb(`Making sure table ${injectedConfig.table} doesn't already exists.`);
+
+  const exists = await connection.schema().hasTable(injectedConfig.table);
+
+  // TODO: Think about a better solution; it shouldn't hav existed in the first place.
```
mesaugat:
```diff
- hav
+ have
```

src/services/configInjection.ts
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * @param {Connection} connection
+ * @param {SyncConfig} config
+ * @returns {Promise<void>}
+ */
+export async function setup(connection: Connection, config: SyncConfig): Promise<void> {
+  const logDb = dbLogger(connection);
+  const { injectedConfig } = config;
+
+  logDb(`Making sure table ${injectedConfig.table} doesn't already exists.`);
```
mesaugat: No need for this log line here as if it did exist beforehand the log down below would be enough.

src/config.ts
```diff
@@ -18,15 +20,43 @@ export async function loadConfig(): Promise<SyncConfig> {
...
+/**
+ * Validate the loaded configuration.
+ *
+ * @param {SyncConfig} config
+ */
+export function validate(config: SyncConfig) {
+  const { injectedConfig } = config;
+
+  // Shouldn't reach under here unless the user has mismatched the value.
+  if (!injectedConfig.table || !injectedConfig.table.trim()) {
```
mesaugat: Let's not allow the user to specify the table name? Because this is a temporary table that the user doesn't need to have access to.

src/services/configInjection.ts
kabirbaidhya: Agreed.

src/config.ts
kabirbaidhya: Makes sense.

kabirbaidhya:
@mesaugat @SafalPandey You can review and merge the example PR into this branch first -> https://github.com/leapfrogtechnology/sync-db/pull/37 and merge both of them together with this PR.

src/services/configInjection.ts
```diff
@@ -0,0 +1,109 @@
+import * as fs from 'fs';
+import * as path from 'path';
+
+import { dbLogger } from '../logger';
+import Mapping from '../domain/Mapping';
+import { expandEnvVarsInMap } from '../util/env';
```
mesaugat: ↕️

src/services/configInjection.ts
```diff
@@ -0,0 +1,109 @@
...
+  return JSON.parse(fs.readFileSync(path.join(__dirname, '../../package.json')).toString());
+}
+
+/**
+ * Gets all the default config / environment variables
+ * that are always available in the injected config table.
+ *
+ * @returns {Mapping<string>}
+ */
+function getDefaultSystemVars(): Mapping<string> {
```
mesaugat: I don't think the doc block and the function are trying to do the same thing?

kabirbaidhya: Actually, it is but the thing is we only have a single param to be passed as a default. It is for now a placeholder for other potential default configs that are to be injected later.

mesaugat: Specially `gets all the environment variables`?

mesaugat: All I see is the version.

kabirbaidhya: For now I've removed everything and have put an empty function as a placeholder for later defaults to inject.

src/services/configInjection.ts
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Setup the table in the database with the injected config.
+ *
+ * @param {Connection} connection
+ * @param {SyncConfig} config
+ * @returns {Promise<void>}
+ */
+export async function setup(connection: Connection, config: SyncConfig): Promise<void> {
+  const logDb = dbLogger(connection);
```
mesaugat: This could simply be `const log`.

src/services/configInjection.ts
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Drop the injected config table.
+ *
+ * @param {Connection} connection
+ * @param {SyncConfig} config
+ * @returns {Promise<void>}
+ */
+export async function cleanup(connection: Connection, config: SyncConfig): Promise<void> {
+  const logDb = dbLogger(connection);
```
mesaugat: Same.

src/constants.ts
```diff
@@ -7,12 +7,16 @@ import SyncConfig from './domain/SyncConfig';
 export const CONFIG_FILENAME = 'sync-db.yml';
 export const CONNECTIONS_FILENAME = 'connections.sync-db.json';
 
+export const INJECTED_CONFIG_TABLE = '__injected_config';
```
mesaugat: Append `sync_db` at the end? `__injected_config_sync_db`

src/services/configInjection.ts
```diff
@@ -0,0 +1,107 @@
...
+  });
+
+  // Inject the configurations into the created table.
+  logDb(`Injecting config into ${INJECTED_CONFIG_TABLE}.`);
+  await connection
+    .getInstance()
+    .insert(values)
+    .into(INJECTED_CONFIG_TABLE);
+
+  logDb(`Inserted configurations: ${values.length}.`);
```
mesaugat: `Inserted ${values.length} configuration values`.

src/constants.ts
kabirbaidhya: I think a prefix would be better - `__sync_db_injected_config` would be better.

mesaugat: Sure.

mesaugat:
@kabirbaidhya Also, why are we not using temporary tables here? It seems like a perfect use case.

https://codingsight.com/introduction-to-temporary-tables-in-sql-server/

kabirbaidhya:
@mesaugat I had thought about this in the first but the problem is - at present we're using multiple connections and/or transaction to run queries even in a single run. 
1. Using local temporary table (`#table`) supports a single session only, it won't span into multiple connections. 
2. Using global temporary table (`##table`) might seem feasible here - but the lifespan might prolong to other sessions too. But this might still be possible someway given we're dropping it at the end ourselves. 

**BUT** now if we are to merge this PR https://github.com/leapfrogtechnology/sync-db/pull/39 first (single transaction per connection) - then no. 1 seem completely possible.

PR #39, opened 2020-02-01
Run synchronize in a single transaction per connection

kabirbaidhya:
### Changes
 - Run the entire process in a single transaction per database connection.
 - Handle errors gracefully for each transaction - even while error occurred in one connection, the process for other connections can still continue.
 - Errors are displayed at the end after once all completed.
 - Better visualization of final result. It displays how many were successful and how many failed. 
 - **Tested** 
    - with single connection
    - with multiple connections both successful and failing. 
    - Experimental versions used for testing v1.0.0-experimental.9-2
 - **Bonus**
   * Display elapsed time for each connection. 

### Screenshots

**1. Successful Output (2 connections, mssql)**
Selection_109

**2. Failing Output (2 connections, mssql)**
Selection_110


src/config.ts
```diff
@@ -47,14 +47,27 @@ export async function resolveConnections(): Promise<ConnectionConfig[]> {
...
   return result;
 }
 
+/**
+ * Get the connection id from the config.
+ *
+ * @param {ConnectionConfig} connectionConfig
+ * @returns {string}
+ */
+export function getConnectionId(connectionConfig: ConnectionConfig): string {
```
kabirbaidhya: Add tests for this.

src/api.ts
```diff
@@ -0,0 +1,106 @@
...
+ */
+function mapToConnectionReferences(connectionList: (ConnectionConfig | Knex)[]): ConnectionReference[] {
+  return connectionList.map(connection => {
+    if (isKnexInstance(connection)) {
+      log(`Received connection instance to database: ${connection.client.config.connection.database}`);
+
+      // TODO: Ask for `id` explicitly in for programmatic API,
+      // when Knex instance is passed directly.
+      // This implies a breaking change with the programmatic API.
+      return { connection, id: undefined };
```
SafalPandey:
```suggestion
      return { connection, id: getConnectionId(getConfig(connection)) };
```

kabirbaidhya:
@mesaugat @SafalPandey @cham11ng This is now fully tested and ready for review. :crossed_fingers:

src/SyncDb.ts
```diff
@@ -20,35 +24,93 @@ class SyncDb extends Command {
...
       process.exit(-1);
     }
   }
+
+  /**
+   * Display the result.
+   *
+   * @param {SyncResult[]} result
+   */
+  async displayResult(result: SyncResult[]) {
```
mesaugat:
```suggestion
  async displayResult(results: SyncResult[]) {
```

src/SyncDb.ts
```diff
@@ -20,35 +24,93 @@ class SyncDb extends Command {
...
+      await printError(e.toString());
+
       process.exit(-1);
     }
   }
+
+  /**
+   * Display the result.
+   *
+   * @param {SyncResult[]} result
```
mesaugat: `results`

src/SyncDb.ts
```diff
@@ -20,35 +24,93 @@ class SyncDb extends Command {
...
+   */
+  async displayResult(result: SyncResult[]) {
+    const failed = result.filter(attempt => !attempt.success);
```
mesaugat: Positive cases are easier to read.
 
```suggestion
    const success = results.filter(result => result.success);
```

src/SyncDb.ts
```diff
@@ -20,35 +24,93 @@ class SyncDb extends Command {
...
+      // Display output.
+      await printLine(`Synchronization successful for ${successfulCount} / ${totalCount} connection(s).`);
+    }
+
+    if (failed.length === 0) {
+      return process.exit(0);
+    }
+
+    // Display all errors.
+    await printLine(`Failed for following ${failedCount} connection(s):\n`);
```
mesaugat:
```suggestion
    await printLine(`Synchronization failed for ${failedCount} connection(s):\n`);
```

src/SyncDb.ts
```diff
@@ -20,35 +24,93 @@ class SyncDb extends Command {
...
+
+    failed.forEach(async (attempt, index) => {
+      await printLine(`${index + 1}) ${attempt.connectionId}`);
+      await printError(attempt.error);
+      await printLine();
+    });
+
+    throw new Error(`Synchronization failed for ${failedCount} / ${totalCount} connections.`);
```
mesaugat: Why throw at the end? Displaying all the errors is not enough?

mesaugat: Why throw at the end? Displaying all the errors is not enough?

src/SyncDb.ts
```diff
@@ -20,35 +24,93 @@ class SyncDb extends Command {
...
+    if (failed.length === 0) {
+      return process.exit(0);
+    }
+
+    // Display all errors.
+    await printLine(`Failed for following ${failedCount} connection(s):\n`);
+
+    failed.forEach(async (attempt, index) => {
+      await printLine(`${index + 1}) ${attempt.connectionId}`);
+      await printError(attempt.error);
```
mesaugat: Ideally the best way is to only show the message for the error and not the whole stack trace. The whole stack trace can be shown in debug mode.

src/api.ts
```diff
@@ -0,0 +1,73 @@
...
+  config: SyncConfig,
+  conn: ConnectionConfig[] | ConnectionConfig | Knex[] | Knex,
+  options?: SyncParams
+): Promise<SyncResult[]> {
+  log('Starting to synchronize.');
+
+  const connectionList = Array.isArray(conn) ? conn : [conn];
+  const connections = mapToConnectionReferences(connectionList);
+  const params = mergeDeepRight(DEFAULT_SYNC_PARAMS, options);
+  const cliEnvironment = process.env.SYNC_DB_CLI === 'true';
```
mesaugat:
```suggestion
  const isCLI = process.env.SYNC_DB_CLI === 'true';
```

src/api.ts
```diff
@@ -0,0 +1,73 @@
...
+  const promises = connections.map(({ connection, id: connectionId }) =>
+    synchronizeDatabase(connection, {
+      config,
+      params,
+      connectionId,
+      cliEnvironment
+    })
+  );
+
+  const result = await Promise.all(promises);
```
mesaugat:
```suggestion
  const results = await Promise.all(promises);
```

src/api.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  log('Finished all');
+
+  return result;
```
mesaugat: .

src/api.ts
```diff
@@ -0,0 +1,73 @@
...
+      config,
+      params,
+      connectionId,
+      cliEnvironment
+    })
+  );
+
+  const result = await Promise.all(promises);
+
+  log('Finished all');
```
mesaugat:
```suggestion
  log('Synchronization completed.');
```

src/SyncDb.ts
kabirbaidhya: Yes, but we need all the failed attempts instead. For successful ones only the counts matter - for error cases we need full error trace. 

```ts
    failed.forEach(async (attempt, index) => {
      await printLine(`${index + 1}) ${attempt.connectionId}`);
      await printError(attempt.error);
      await printLine();
    });
```

src/api.ts
```diff
@@ -0,0 +1,73 @@
...
+function mapToConnectionReferences(connectionList: (ConnectionConfig | Knex)[]): ConnectionReference[] {
+  return connectionList
+    .map(connection => {
+      if (isKnexInstance(connection)) {
+        log(`Received connection instance to database: ${connection.client.config.connection.database}`);
+
+        // TODO: Ask for `id` explicitly in for programmatic API,
+        // when Knex instance is passed directly.
+        // This implies a breaking change with the programmatic API.
+        return { connection, id: undefined };
```
mesaugat: Not sure if I understand correctly but do we need the second `map`?

```js
return { connection, id: getConnectionId(getConfig(connection)) }
```

src/domain/SyncContext.ts
```diff
@@ -0,0 +1,14 @@
+import SyncConfig from './SyncConfig';
+import SyncParams from './SyncParams';
+
+/**
+ * Synchronize context parameters for the current database connection.
+ */
+interface SyncContext {
+  config: SyncConfig;
+  connectionId: string;
+  cliEnvironment: boolean;
```
mesaugat: .

src/services/dbSyncer.ts
```diff
@@ -0,0 +1,130 @@
+import * as Knex from 'knex';
```
mesaugat: Filename could still be `dbSync`. Haha `dbSyncer`?

kabirbaidhya: Naming is always difficult :slightly_smiling_face:

src/SyncDb.ts
SafalPandey: We should throw the error to make sure the program exits with error code other than 0.

src/api.ts
kabirbaidhya: Both are user provided values so we'll never be sure if we get the `id` out in both cases. Here we won't get it anyway so it's `undefined` to explicitly say so. 

If we do it here - then we'll need to do it in two places - that's why I moved it to a second map - this ensures all the connections has id (or has a fallback id if not provided)

src/api.ts
```diff
@@ -0,0 +1,73 @@
...
+      }
+
+      log(`Creating a connection to database: ${connection.host}/${connection.database}`);
+
+      return { connection: createInstance(connection), id: connection.id };
```
kabirbaidhya: @mesaugat There's no assurance that `id` has a value. This is just a typecheck not a run-time assurance.

src/SyncDb.ts
kabirbaidhya: Yes, or else we'll need to exit non-zero in multiple places which isn't good either.

src/SyncDb.ts
kabirbaidhya: Make sense. I'll send out the stack trace in `DEBUG` mode.

kabirbaidhya:
@mesaugat This PR is a small one - which could be merged first and I can rebase it here before making changes in that utility.

src/SyncDb.ts
mesaugat: So `process.exit(1)`

src/api.ts
mesaugat: I still feel doing the same in two place is better than using a `map`. Don't need to be DRY everywhere and is a lot easier to understand. 🤷‍♂

src/SyncDb.ts
kabirbaidhya: It's not a good idea of exit in multiple places - throwing exceptions is the right way to raise errors. We just need to ensure that there's a root handler which catches all of them.

mesaugat: Cool. We already have a try catch in place. 

My only issue is why does a function that simply displays a summary or a result needs to throw any kind of error.

kabirbaidhya: There's a single place where all exceptions are caught and a non zero exit code is thrown in case of errors.

kabirbaidhya: Good point. This doesn't sound right with the display result function as it says.

src/SyncDb.ts
```diff
@@ -20,35 +24,98 @@ class SyncDb extends Command {
...
   }
+
+  /**
+   * Display the results.
+   *
+   * @param {SyncResult[]} results
+   */
+  async displayResult(results: SyncResult[]) {
+    const totalCount = results.length;
+    const failedAttempts = results.filter(attempt => !attempt.success);
```
mesaugat:
```suggestion
    const failedAttempts = results.filter(result => !result.success);
```

src/SyncDb.ts
mesaugat: The problem is that `displayResult` did not fail but `synchronize` or something internally did. Right now display result is trying to say that something failed upstream thus I should fail. When everything is well, `displayResult` does what it is supposed to do. It prints the result and that is what it should do.

The throw should happen after `displayResult`.

```js
      const results = await synchronize(config, connections, params);

      await this.displayResult(results);

      const fail = results.find(result => result.success === false)
      
      if (fail) {
         throw new Error('...');
      }
    } catch (e) {
      // Send verbose error with stack trace to debug logs.
      log(e);

      await printError(e.toString());

      process.exit(-1);
    }
```

src/SyncDb.ts
```diff
@@ -20,35 +24,104 @@ class SyncDb extends Command {
...
+      const { synchronize } = await import('./api');
+
+      await printLine('Synchronizing...\n');
 
-      await synchronize(config, connections, params);
-      process.exit(0);
+      const results = await synchronize(config, connections, params);
+      const { totalCount, failedCount } = await this.processResults(results);
+
+      if (failedCount > 0) {
```
mesaugat: Haha, I made a similar comment and you pushed it as well.

mesaugat: https://github.com/leapfrogtechnology/sync-db/pull/39#discussion_r384325669

PR #41, opened 2020-02-03
Avoid vendor specific defaults in the config

kabirbaidhya:
Port number 1433 is for mssql only; sync-db is vendor agnostic - at least it supports mssql and pg at the moment.


mesaugat:
O.o Nice catch. So into MSSQL that I forgot that there are other databases that exist.

PR #50, opened 2020-03-09
Support resolution of local project dependencies when CLI is invoked in development mode

kabirbaidhya:
Make the main CLI process resolve the local project dependencies (from the PWD) even if the CLI is not installed locally and has been invoked from a different path.

- Add a new script `bin/run-dev.sh` for running the CLI (in development mode) - this provide `NODE_PATH` including the node_modules path of the PWD to the node process; allowing the CLI's require calls to fallback to the local packages from the PWD if needed to be resolved.
   
   Read more on how it [works](https://gist.github.com/branneman/8048520#51-up-front) [here](https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders).

### Problem
So far, when the CLI is invoked from a different context other than what has been installed inside your local project dependencies (via local `package.json`).

For instance: While developing, if you want to run it in a project to verify after you make some changes, you'll get this error:
image

And the problem here is - the CLI can't resolve the local dependencies of the local project as the project directory for the CLI and the execution context (PWD) are different.

### Benefits
- ~Both global and local installation just work work as expected.~
- It allows you to test `sync-db` in the examples or your local projects during development w/o the need to publish a new version.

    See it working here for a local example project.
    image


bin/run
```diff
@@ -1,14 +1,24 @@
...
+#!/bin/sh
 
-const fs = require('fs');
-const path = require('path');
-const project = path.join(__dirname, '../tsconfig.json');
-const dev = fs.existsSync(project);
+# Get the local node_modules path;
+# if NODE_PATH is already set, append it to the existing value.
+if [ -n "$NODE_PATH" ]; then
+  UPDATED_PATH="${NODE_PATH}:${PWD}/node_modules"
```
mesaugat: What is the ":" colon doing in the path?

bin/run
```diff
@@ -1,14 +1,24 @@
...
+#
+# Use-case:
+# The main reason for this is to allow the CLI's node process
+# to be able to resolve local project dependencies;
+# this is important when the CLI is installed globally and
+# the local project installation has some of the packages that the CLI needs.
+# For eg: if the local project has mssql package installed and the CLI is invoked globally.
+#
+# This allows the CLI's require calls to fallback to the local dependencies if needed.
+NODE_PATH="${UPDATED_PATH}" node $BASEDIR/run.js
```
mesaugat: So basically what we are saying is we don't care about Windows. This doesn't work in windows.

mesaugat: Do we need to make this a global CLI tool? I think we should opt out of that because we depend on packages like mssql, psql and knex.

bin/run
kabirbaidhya: Appending the new value to the existing list of values. It works the same as `$PATH`. 
Something like this:
```
/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/lib/nodejs/node-v12.14.1-linux-x64/bin:/source/node_modules/.bin
```

bin/run
kabirbaidhya: Makes sense.

kabirbaidhya: Yes, good catch. Need to tweak the `run.cmd` a bit for this to make it work.

kabirbaidhya: @mesaugat This is what this PR does. The CLI is not tied with the local dependencies - be it global or local, it's the user's choice. It just supports it both.

When installed globally, it resolves the local dependencies from the local project directory normally. So, there's no dependency in between them. 

Other two reasons are:
 - Right now, it's difficult to test. When you make any change you cannot test it locally unless you publish it as a package and install it in your local project - this is painful and slow. But #50 makes it seamless - you can do `/path/to/local/bin/run` from your project.
- Secondly, you could publish a docker image for sync-db - so the user doesn't even need to install sync-db in their local - it would be like a binary. They'll just need their local deps like mssql & configurations; not even knex is required.

mesaugat: @kabirbaidhya I get all of that and I feel this is not required. This is true for any kind of package and we don't need to publish it. We just need do `yarn pack` or `npm pack` and create a package locally. Maybe let the CI do this. Then, we need to install that package from the locally zipped tar to test.

mesaugat: > A binary should mostly be self-sufficient. 

`sync-db` always needs a SQL client package to run. So creating a binary that doesn't do anything by itself doesn't make any sense to me. Either we have to package the individual client drivers with the binary or stop thinking about running it globally.

"Someone installs sync-db globally and then wants to run it in their project. Now, they need to install the SQL client and add it to their `package.json` as a local dependency" or maybe install the driver globally as well." 

There are simply too many confusions with the global approach. A simple install these dependencies and run it locally like every other JavaScript package is the best approach in my opinion.

kabirbaidhya: @mesaugat I have a different perspective on this. 
> A binary should mostly be self-sufficient.

Actually, the binary is self-dependent. It is just that the project needs to have a vendor specific connector to talk to the corresponding DB they use; but that isn't a part of CLI itself. And, they'll need configuration as per their project's source - that is project specific too. So, actually you can think of the CLI as a self-sufficient binary as one global installation would work with multiple projects configured differently.

But, I also agree to the point that global installation might be confusing for users and might be harder for us to push BC breaking changes in longer run. 

> We just need do yarn pack or npm pack and create a package locally. 

This is not sustainable; testing locally is really painful w/ this approach and this won't work as it's very troublesome at least for sync-db. Because there are various things you'll need to test trying out in different projects before you can push.

So, net-net, we'll need something like this:
1. don't encourage global install -- Yes
2. have this as a separate script (won't be published) to run locally eg: `run.sh`

Thoughts?

mesaugat: Cool.

PR #51, opened 2020-03-09
Migrations support

kabirbaidhya:
Tons of changes -- too many to be mentioned. Expect a clean write up once all complete.

PR #58, opened 2020-03-24
Transition to multi commands CLI

kabirbaidhya:
**BREAKING CHANGE**
- Requires node >= 10.x (Merge https://github.com/leapfrogtechnology/sync-db/pull/60 first)
- Need to transition to multi-commands CLI now as we're anticipating the migrations feature soon as per #51. This PR changes the CLI to allow that.
- You'll need to run `synchronize` command now to run the existing functionality:
   ```bash
   $ sync-db synchronize
   ```
   Previously, running just `sync-db` would have worked.
- Add CLI tests   
- **Preview**
   image

**Why?**
This sets the ground for upcoming features we're going to have. We're going to need the following kinds of commands sooner:
```bash
- synchronize
- migrate:latest
- migrate:rollback
- migrate:list
- config:validate
- init
```

### DIFF
--- a/src/commands/synchronize.ts
+++ b/src/commands/synchronize.ts
@@ -10,9 +10,9 @@ import ExecutionContext from '../domain/ExecutionContext';
 import { loadConfig, resolveConnections } from '../config';
 
 /**
- * SyncDB CLI handler.
+ * Synchronize command handler.
  */
-class SyncDb extends Command {
+class Synchronize extends Command {
   static description = 'Synchronize database';
 
   /**
@@ -50,7 +50,7 @@ class SyncDb extends Command {
    * @returns {Promise<void>}
    */
   async run(): Promise<void> {
-    const { flags: parsedFlags } = this.parse(SyncDb);
+    const { flags: parsedFlags } = this.parse(Synchronize);
     const params = this.getSyncParams({ ...parsedFlags });
 
     try {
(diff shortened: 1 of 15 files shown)


src/commands/synchronize.ts
```diff
@@ -1,18 +1,18 @@
...
 
 /**
- * SyncDB CLI handler.
+ * Synchronize command handler.
  */
-class SyncDb extends Command {
+class Synchronize extends Command {
   static description = 'Synchronize database';
 
   /**
```
SafalPandey, line 18: Perhaps `generate-connections` can be a command now instead of a flag for `synchronize`.

I think this looks a bit weird:
```bash
yarn sync-db synchronize --generate-conections
```

This keeps it simple:
```bash
yarn sync-db generate-connections
```

kabirbaidhya: We'll remove generate-connections now. It's not useful as the env vars being used by this flag can directly be used without a connections file if given. 

So, it would be simpler for the users to know less on this. 

Thanks for pointing it out thought, I'll put a deprecation log for it.

kabirbaidhya:
Merging this to `experimental` to proceed the further work.

kabirbaidhya:
Keep reviewing and suggesting on the main PR here -> https://github.com/leapfrogtechnology/sync-db/pull/62

PR #59, opened 2020-03-27
Add `enableArithAbort` flag in sync_db generate connection util

cham11ng:
Add `enableArithAbort` flag in sync_db generate connection util


kabirbaidhya:
@cham11ng I'm closing this, because this is a vendor specific configuration which shouldn't get here in sync-db. It's a configuration used by node-mssql which is something users can solve by following the warning and putting this in their `connections.sync-db.json` file. We cannot put client specific config in our base.

Plus, if knex fixes this it's good for us, we'll bump our version to it and/or if node-mssql publishes a new version, we'll move to that too. So, we won't need this temporary change here.

PR #42, opened 2020-02-03
(WIP) Ability to statically validate config and source sql

kabirbaidhya:
- Ability to statically validate the configuration and the source sql files
- Add tests

PR #62, opened 2020-04-07
Breaking changes - migrations, multi-commands and more experimental changes

kabirbaidhya:
- [x] Migrations support (SQL based, inspired by [go-migrate](https://github.com/golang-migrate/migrate) but fully based on [Knex's API](http://knexjs.org/#Migrations-API))
- [x] Multi-commands CLI
- [x] Other significant updates / several breaking changes
- [x] Custom connection resolver support (#69)

---

**Screenshots**
image
image
image


kabirbaidhya:
@mesaugat @SafalPandey @cham11ng Heads up - tons of changes. 
You may skim through the code and leave comments but this is still work-in-progress / experimental; I'll share a tutorial for this once finished and we can go for final review and beta test.

mesaugat:
@kabirbaidhya Are we not using this version already? Should we merge this? And what about the other PR #42?

kabirbaidhya:
@mesaugat Yes, we are. The beta/experimental versions are released but we have been using it in production already since few months and its kind of stable. 

You can ignore #42 change right now - it's completely independent and could be worked on later after this merge.

src/domain/Configuration.ts
```diff
@@ -0,0 +1,26 @@
...
+    post_sync: string[];
+  };
+  injectedConfig: {
+    vars: Mapping<string>;
+  };
+  migration: {
+    directory: string;
+    tableName: string;
+    // TODO: Only 'sql' is supported sourceType now. JS will be supported later.
+    sourceType: 'sql' | 'javascript';
```
kabirbaidhya, line 22: @mesaugat This `javascript` based migration file support is currently missing. We can merge this and add that support soon in the next release. 

All we need to add here, is an ability to import { up, down } functions from the migration files. It won't be that hard, given all the groundwork is already done here.

mesaugat: Nice.

mesaugat (approved):
Merging, didn't look at all the details. We have been already using this in production.

PR #65, opened 2020-04-26
Make connections config consistent with Knex's config (breaking change)

kabirbaidhya:
**BREAKING CHANGE**
- Connection config is made identical to the Knex's config. The only difference is we use only handful of those `client` and `connection` and we add `id` attribute for our internal purpose.
- Being consistent with Knex is important because we want to support the things Knex add out of the box, as much as possible.
- Programmatic API usage will break too - it expects type `ConnectionConfig | ConnectionConfig[]`.

**Checklist**
- [x] Update and test node-app-mssql example
- [ ] Update and test programmatic usage example
- [ ] Update connection id auto-generation logic when not given - making it numeric might be the only way there.

### DIFF
--- a/examples/node-app-mssql/connections.sync-db.json.docker
+++ b/examples/node-app-mssql/connections.sync-db.json.docker
@@ -2,12 +2,14 @@
   "connections": [
     {
       "id": "testdb1",
-      "host": "mssql",
-      "port": 1433,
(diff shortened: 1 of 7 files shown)


kabirbaidhya:
Remaining TODO items and tests to be done in the subsequent PRs.

PR #68, opened 2020-05-17
Introduce make command to generate migration files

kabirbaidhya:
### The make command
* Introduce `make` command to generate migration files.
* If it receives an argument in the form `create_<something>_table` then it will use the create table template for the file otherwise blank migration files are generated.
* In future it could support creating other DB objects like views, functions, etc with an optional option `--type` given

### Other chnages
* Add ts utili function, few more tests
* Remove CLI tests from test coverage - it's not needed and it took very long time for the tests.

### Preview
```
$ sync-db make create_test_table
```
image

```
$ sync-db make somethingelse
```
image

### DIFF
--- a/src/service/fileMaker.ts
+++ b/src/service/fileMaker.ts
@@ -1,6 +1,7 @@
 import * as path from 'path';
 
 import * as fs from '../util/fs';
+import { getTimestampString } from '../util/ts';
 import Configuration from '../domain/Configuration';
 import { getMigrationPath } from '../migration/service/knexMigrator';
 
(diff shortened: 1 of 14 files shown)


src/util/string.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Interpolate a given template string by filling the placeholders with the params.
+ *
+ * Placeholder syntax:
+ *  {{name}}
+ *
+ * @example
+ * interpolate('<div>>{{text}}</div>', {text: 'Hello World!'})
```
silwalanish, line 8:
```suggestion
 * interpolate('<div>{{text}}</div>', {text: 'Hello World!'})
```

src/service/fileMaker.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    createUpTemplate = await fs
+      .read(path.join(MIGRATION_TEMPLATE_PATH, 'create_up.sql'))
+      .then(template => interpolate(template, { table }));
+    createDownTemplate = await fs
+      .read(path.join(MIGRATION_TEMPLATE_PATH, 'create_down.sql'))
+      .then(template => interpolate(template, { table }));
+  }
+
+  await fs.write(upFilename, createUpTemplate);
```
silwalanish, line 58: I think we should check if the migration doesn't exists already. And warn or throw if it does already?

silwalanish: In second thought, although the timestamp will prevent from overriding. Having multiple migration with same name maybe confusing but let's leave that to the developers.

kabirbaidhya: There is very rare chance. It won't happen because it's prefixed with timestamp that changes every second. So, no need.

PR #93, opened 2021-03-05
Automate sync-db release

### LINKED ISSUE #89: Remove support for Node 10
Node 10 has already dropped active support and it's security support is ending in a few months time. 

endoflife date_nodejs

* Update `package.json` engine version to use Node 12 LTS and above
* Remove Node 10 from `.travis.yml` and include Node 12, 14
* Upgrade packages that don't support Node 12

samirsilwal:
## Changes
- Updated release script to automate git tag 
- updated travis.yml to include deploy stage
- Added compare_and_release function which finds value for `NEXT` value extracting change in package.json
- deploy step runs only on master branch and publish only if version of package is changed

closes #89 

**NOTE:- Change log generation should be done manually.**


.travis.yml
```diff
@@ -12,3 +19,13 @@ before_script:
 script:
   - yarn test:coverage
   - yarn codecov
+  - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo "$TRAVIS_BRANCH"; else echo "$TRAVIS_PULL_REQUEST_BRANCH"; fi)
+  - if [ "$BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then ./release.sh compare_and_release; fi
+
+deploy:
+  provider: npm
+  email: '$NPM_EMAIL'
```
mesaugat, line 27: Is this mandatory?

.travis.yml
```diff
@@ -12,3 +19,13 @@ before_script:
...
+  api_key: '$NPM_API_TOKEN'
```
mesaugat, line 28: Let's make it `NPM_TOKEN` for consistency.

```suggestion
  api_key: '$NPM_TOKEN'
```

.travis.yml
```diff
@@ -2,7 +2,14 @@ language: node_js
 
 node_js:
   - node
```
mesaugat, line 4: While you are at it. Add Node 12 and Node 14 as well.

.travis.yml
```diff
@@ -12,3 +19,13 @@ before_script:
...
+  - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo "$TRAVIS_BRANCH"; else echo "$TRAVIS_PULL_REQUEST_BRANCH"; fi)
+  - if [ "$BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then ./release.sh compare_and_release; fi
+
+deploy:
+  provider: npm
+  email: '$NPM_EMAIL'
+  api_key: '$NPM_TOKEN'
+  skip_cleanup: true
+  on:
+    branch: master
```
mesaugat, line 31: Looks like we will need to release when we push tags?

.travis.yml
mesaugat: Is this because publish will happen multiple times if we specify multiple node versions here?

.travis.yml
```diff
@@ -12,3 +24,17 @@ before_script:
 script:
   - yarn test:coverage
   - yarn codecov
+  - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo "$TRAVIS_BRANCH"; else echo "$TRAVIS_PULL_REQUEST_BRANCH"; fi)
+  - if [ "$BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then ./release.sh compare_and_release; fi
+
+jobs:
+  include:
+    - stage: test
```
mesaugat, line 32: stage: test?

samirsilwal: This is default stage name. I was trying to test some conditioning but it turns out it would work same anyway with any stage name. reverting to deploy

release.sh
```diff
@@ -25,28 +29,54 @@ changelog() {
...
+  fi
+}
+
+compare_and_release() {
+  ## Compare the package.json file from two recent commits to master branch and export value to NEXT variable
+  ## if differs else NEXT=false
 
-  # Update version in the following files
-  sed -i "s/\(\"version\":\s*\"\)[^\"]*\(\"\)/\1${VERSION}\2/g" package.json
+  previous_commit_hash=$(git rev-parse @~)
```
mesaugat, line 53: Isn't this the same? We are trying to get the last commit in master?

```suggestion
  previous_commit_hash=$(git rev-parse HEAD)
```

release.sh
```diff
@@ -25,28 +29,54 @@ changelog() {
...
+  printfln "New package version: ${new_version}"
+
+  NEXT=false
+
+  if [ "$old_version" != "$new_version" ]; then
+    printfln "Publishing changes to npm with version: ${new_version}."
+    NEXT=${new_version}
+  fi
+
+  printfln "Value of NEXT is: ${NEXT}."
```
mesaugat, line 75: We might not need this log?

release.sh
samirsilwal: Not last but second last. @ -> last @~ -> second last. when PR is merged lst commit contains changed version

release.sh
```diff
@@ -25,28 +29,54 @@ changelog() {
 }
 
 bump() {
-  # Bump package version and generate changelog
-  VERSION="${NEXT/v/}"
 
-  echo "Bump version to ${VERSION}"
+  if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+      printfln "Exiting for pull request without versioning."
```
mesaugat, line 34: This doesn't exit the script. I'm confused. Should we exit here? It only prints.

samirsilwal: Oh will fix it

release.sh
mesaugat: Cool. Can you add a comment here that says.

```
# Get the second last commit from the master branch after merge.
```

release.sh
mesaugat, line 34:
```suggestion
      printfln "Skipping release for pull request"
```

release.sh
```diff
@@ -25,28 +29,54 @@ changelog() {
...
+  git checkout master
+
+  new_version=$(cat package.json | jq -r ".version")
+
+  printfln "New package version: ${new_version}"
+
+  NEXT=false
+
+  if [ "$old_version" != "$new_version" ]; then
+    printfln "Publishing changes to npm with version: ${new_version}."
```
mesaugat, line 73: Remove thoplo at the end because the version will look confusing.

```suggestion
    printfln "Publishing changes to npm with version: ${new_version}"
```

samirsilwal: ok :smile:

.travis.yml
```diff
@@ -2,7 +2,19 @@ language: node_js
 
 node_js:
   - node
-  - 10
+  - 12
+  - 14
```
mesaugat, line 6: Slight OCD. Lower to higher versions. 

@samirsilwal Can you also change `"node": ">= 10.16.0"` in package.json to `"node": ">= 12.13.0"`

```suggestion
  - 14
  - 12
```

mesaugat (approved):
LGTM. Let's create a new PR tomorrow that will bump the `package.json` version to `v1.0.0-beta.6`.

samirsilwal:
> LGTM. Let's create a new PR tomorrow that will bump the `package.json` version to `v1.0.0-beta.6`.

OK will do that

.travis.yml
```diff
@@ -12,3 +24,17 @@ before_script:
 script:
   - yarn test:coverage
   - yarn codecov
+  - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo "$TRAVIS_BRANCH"; else echo "$TRAVIS_PULL_REQUEST_BRANCH"; fi)
+  - if [ "$BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then ./release.sh compare_and_release; fi
```
mesaugat, line 28: @samirsilwal I just figured that we are already checking this here so we can remove these checks from `release.sh`? That way we will be able to release.sh locally as well.

PR #94, opened 2021-03-08
Integrate dry run feature by introducing `--dry-run` flag

cham11ng:
Integrate dry run feature by introducing `dryRun` flag

Problem:

- We are not able to test whether database synchronization will fail/pass before the actual deployment

Fix:

- Added a new command/flag called dry-run in sync-db to check database synchronization.
- The transaction fails in case of a dry run as it will only check if database objects can be synchronized or not.
- Plan: sync-db dry run should happen in pull requests to catch issues before deployment.

Logs:

l Pass | Fail |
|--|--|
|image| image|
|image|image|
|image|image|
|image|image|

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -41,6 +41,7 @@ export async function synchronize(
     'skip-migration': false,
     ...options
   };
+  const { onStarted, ...invokeParams } = params;
 
   // TODO: Need to preload the SQL source code under this step.
   const { knexMigrationConfig } = await init.prepare(config, {
@@ -62,7 +63,7 @@ export async function synchronize(
               config,
               connectionId: connection.id,
               knexMigrationConfig: knexMigrationConfig(connection.id),
-              params: { ...params, onSuccess: params.onMigrationSuccess, onFailed: params.onMigrationFailed }
+              params: { ...invokeParams, onSuccess: params.onMigrationSuccess, onFailed: params.onMigrationFailed }
             })
         }),
       params['dry-run']
--- a/src/api.ts
+++ b/src/api.ts
@@ -41,7 +41,7 @@ export async function synchronize(
     'skip-migration': false,
     ...options
   };
-  const { onStarted, ...invokeParams } = params;
+  const { onStarted: _, ...invokeParams } = params;
 
   // TODO: Need to preload the SQL source code under this step.
   const { knexMigrationConfig } = await init.prepare(config, {
--- a/src/api.ts
+++ b/src/api.ts
@@ -50,19 +50,22 @@ export async function synchronize(
 
   const connections = filterConnectionsAsRequired(mapToConnectionReferences(conn), params.only);
   const processes = connections.map(connection => () =>
-    withTransaction(connection, trx =>
-      runSynchronize(trx, {
-        config,
-        params,
-        connectionId: connection.id,
-        migrateFunc: t =>
-          invokeMigrationApi(t, KnexMigrationAPI.MIGRATE_LATEST, {
-            config,
-            connectionId: connection.id,
-            knexMigrationConfig: knexMigrationConfig(connection.id),
-            params: { ...params, onSuccess: params.onMigrationSuccess, onFailed: params.onMigrationFailed }
-          })
-      })
+    withTransaction(
+      connection,
+      trx =>
+        runSynchronize(trx, {
+          config,
+          params,
+          connectionId: connection.id,
+          migrateFunc: t =>
+            invokeMigrationApi(t, KnexMigrationAPI.MIGRATE_LATEST, {
+              config,
+              connectionId: connection.id,
+              knexMigrationConfig: knexMigrationConfig(connection.id),
+              params: { ...params, onSuccess: params.onMigrationSuccess, onFailed: params.onMigrationFailed }
+            })
+        }),
+      params['dry-run']
     )
   );
 
@@ -93,12 +96,15 @@ export async function prune(
 
   const connections = filterConnectionsAsRequired(mapToConnectionReferences(conn), params.only);
   const processes = connections.map(connection => () =>
-    withTransaction(connection, trx =>
-      runPrune(trx, {
-        config,
-        params,
-        connectionId: connection.id
-      })
+    withTransaction(
+      connection,
+      trx =>
+        runPrune(trx, {
+          config,
+          params,
+          connectionId: connection.id
+        }),
+      params['dry-run']
     )
   );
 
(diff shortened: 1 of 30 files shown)


src/util/db.ts
```diff
@@ -60,17 +60,32 @@ export function createInstance(config: ConnectionConfig): Knex {
...
  * Run a callback function with in a transaction.
+ * If `dryRun` is true, transaction will not be committed.
  *
  * @param {ConnectionReference} db
  * @param {(trx: Knex.Transaction) => Promise<T>} callback
+ * @param {boolean} dryRun?
  * @returns {Promise<T>}
  */
-export function withTransaction<T>(
+export async function withTransaction<T>(
```
cham11ng, line 70: Main logic.

cham11ng: cc @SafalPandey @mesaugat @samirsilwal

src/util/db.ts
```diff
@@ -60,17 +60,32 @@ export function createInstance(config: ConnectionConfig): Knex {
...
   const dbLog = dbLogger(db.id);
 
+  if (dryRun) {
+    dbLog('BEGIN: Dry Run transaction');
+
+    const trx = await db.connection.transaction();
+    const res = await callback(trx);
+
+    dbLog('END: DRY RUN transaction');
+    dbLog('END: DRY RUN trx.isCompleted():', trx.isCompleted());
```
SafalPandey, line 84: Shouldn't the transaction be rolled back here? Not sure I understand what's happening here.
Would be nice to have some comments explaining how actual DB changes are being prevented.

src/api.ts
```diff
@@ -41,6 +41,7 @@ export async function synchronize(
     'skip-migration': false,
     ...options
   };
+  const { onStarted, ...invokeParams } = params;
```
SafalPandey, line 44: No ts-lint `unused-variable` errors?

```suggestion
  const { onStarted: _, ...invokeParams } = params;
```

src/commands/migrate-latest.ts
```diff
@@ -21,6 +22,15 @@ class MigrateLatest extends Command {
     })
   };
 
+  /**
+   * Started event handler.
+   */
+  onStarted = async (result: OperationResult) => {
+    await printLine(bold(` ▸ ${result.connectionId}`));
+    await printLine(bold(' ▸ DRY RUN STARTED'));
```
SafalPandey, line 30: Think these two lines should look a bit different as they're trying to represent different things.

Would be better if we can use something other than ` ▸` for dry run log.
This will help easily distinguish `connection id` from the `process`.

```suggestion
    await printLine(bold(` ▸ ${result.connectionId}`));
    await printLine(' • DRY RUN STARTED');
```

kabirbaidhya:
Nice. This is feature to have. I remember we had it in our wishlist for a longtime.

src/api.ts
```diff
@@ -50,19 +51,22 @@ export async function synchronize(
...
+          connectionId: connection.id,
+          migrateFunc: t =>
+            invokeMigrationApi(t, KnexMigrationAPI.MIGRATE_LATEST, {
+              config,
+              connectionId: connection.id,
+              knexMigrationConfig: knexMigrationConfig(connection.id),
+              params: { ...invokeParams, onSuccess: params.onMigrationSuccess, onFailed: params.onMigrationFailed }
+            })
+        }),
+      params['dry-run']
```
kabirbaidhya, line 69: Why are you passing the dry-run params separately here. The whole params object has been already passed above - shouldn't you use that?

src/api.ts
```diff
@@ -93,12 +97,15 @@ export async function prune(
...
-      })
+    withTransaction(
+      connection,
+      trx =>
+        runPrune(trx, {
+          config,
+          params,
+          connectionId: connection.id
+        }),
+      params['dry-run']
```
kabirbaidhya, line 108: same everywhere else.

src/api.ts
kabirbaidhya: Understood now. Ignore this comment.

kabirbaidhya:
One suggestion, `--dry-run` should also print the list of all the statements that would have been run (possibly in the stdout). 
This way it could be used to test everything. Alternatively, if we don't want to be that verbose every time we do it, then we may add an additional flag to be used in conjunction to print out such as `--dry-run --print-sql`.

cc @mesaugat

mesaugat:
@kabirbaidhya Sounds good. It might pollute the terminal big time. I'll create an issue for it so that we can take care of it later.

mesaugat:
@cham11ng CI fail. https://travis-ci.com/github/leapfrogtechnology/sync-db/jobs/489254955

cham11ng:
@kabirbaidhya @mesaugat We can still view debug log to view all the queries.

```
DEBUG=knex:*
```

mesaugat:
Can we add a screenshot dry run failure?

src/commands/migrate-latest.ts
```diff
@@ -63,21 +71,29 @@ class MigrateLatest extends Command {
    */
   async run(): Promise<void> {
     const { flags: parsedFlags } = this.parse(MigrateLatest);
+    const isDryRun = parsedFlags['dry-run'];
     const config = await loadConfig();
     const connections = await resolveConnections(config, parsedFlags['connection-resolver']);
 
+    if (isDryRun) await printLine('• DRY RUN STARTED\n');
```
mesaugat, line 78: Can make this yellow.

src/commands/synchronize.ts
```diff
@@ -149,6 +152,8 @@ class Synchronize extends Command {
       const { totalCount, failedCount, successfulCount } = await this.processResults(results);
 
       if (successfulCount > 0) {
+        if (isDryRun) await printLine(green('[✓] DRY RUN SUCCESS\n'));
```
mesaugat, line 155: Let's print this at the end for consistency.

src/commands/synchronize.ts
```diff
@@ -161,6 +166,8 @@ class Synchronize extends Command {
         return process.exit(0);
       }
 
+      if (isDryRun) await printLine(red('[✖] DRY RUN FAILED\n'));
```
mesaugat, line 169: Let's print success and failure here.

src/commands/prune.ts
```diff
@@ -43,21 +50,29 @@ class Prune extends Command {
    */
   async run(): Promise<void> {
     const { flags: parsedFlags } = this.parse(Prune);
+    const isDryRun = parsedFlags['dry-run'];
     const config = await loadConfig();
     const connections = await resolveConnections(config, parsedFlags['connection-resolver']);
 
+    if (isDryRun) await printLine('• DRY RUN STARTED\n');
+
     const results = await prune(config, connections, {
```
mesaugat, line 59: Maybe I don't understand but why do we need dry run logs here. We only need it at the top level? Just curious can you put a screenshot of migrate-rollback and prune as well. Thanks.

src/util/db.ts
```diff
@@ -60,17 +60,33 @@ export function createInstance(config: ConnectionConfig): Knex {
...
 
+  if (dryRun) {
+    dbLog('BEGIN: Dry Run transaction');
+
+    const trx = await db.connection.transaction();
+    const res = await callback(trx);
+
+    trx.rollback();
+
+    dbLog('END: Dry Run transaction rolled-back successfully');
```
mesaugat, line 85:
```suggestion
    dbLog('END: Dry Run transaction rolled back successfully');
```

src/commands/prune.ts
cham11ng: image

cc @mesaugat

mesaugat (approved):
Minor queries from my side. Looks good overall. This is going to be so handy. 🎉

src/commands/prune.ts
mesaugat: Nice. Makes sense now. 👍

PR #98, opened 2021-03-09
Update script to auto generate changelog

samirsilwal:
## Changes 
- Updated `release.sh` to auto generate changelog once the package is tagged

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -18,11 +18,13 @@ changelog() {
 
   echo "Generating changelog upto version: $NEXT"
   github_changelog_generator \
+    --user="leapfrogtechnology" \
+    --project="sync-db" \
+    --token="$GITHUB_TOKEN" \
     --no-verbose \
     --pr-label "**Changes**" \
     --bugs-label "**Bug Fixes**" \
     --issues-label "**Closed Issues**" \
-    --issue-line-labels=ALL \
     --future-release="$NEXT" \
     --release-branch=master \
     --exclude-labels=unnecessary,duplicate,question,invalid,wontfix
@@ -69,6 +71,13 @@ compare_and_release() {
 
   if [ -n "$NEXT" ] && [ "$NEXT" != "false" ]; then
     bump
+    changelog
+
+    git add CHANGELOG.md
+    git commit -v --edit -m "${NEXT} Release :tada: :fireworks: :bell:"
+    git push origin master
+
+    printfln "Updated changelog."
   fi
 }
 
--- a/release.sh
+++ b/release.sh
@@ -76,9 +76,7 @@ compare_and_release() {
     git config --global user.name "Travis CI"
 
     git add CHANGELOG.md
-    git commit -v --edit -m "${NEXT} Release :tada: :fireworks: :bell:
-    [skip ci]
-    "
+    git commit -v --edit -m "${NEXT} Release :tada: :fireworks: :bell:" -m "[skip ci]"
 
     git remote rm origin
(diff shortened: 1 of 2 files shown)


release.sh
```diff
@@ -18,11 +18,13 @@ changelog() {
...
   echo "Generating changelog upto version: $NEXT"
   github_changelog_generator \
+    --user="leapfrogtechnology" \
+    --project="sync-db" \
+    --token="$GITHUB_TOKEN" \
     --no-verbose \
     --pr-label "**Changes**" \
     --bugs-label "**Bug Fixes**" \
     --issues-label "**Closed Issues**" \
-    --issue-line-labels=ALL \
```
samirsilwal, line 25: @kabirbaidhya @mesaugat i was getting this error locally 
Screenshot from 2021-03-09 17-37-29
says notsuchmethod so i removed it. does it affect anything ?

release.sh
```diff
@@ -69,6 +71,13 @@ compare_and_release() {
 
   if [ -n "$NEXT" ] && [ "$NEXT" != "false" ]; then
     bump
+    changelog
+
+    git add CHANGELOG.md
+    git commit -v --edit -m "${NEXT} Release :tada: :fireworks: :bell:"
+    git push origin master
+
+    printfln "Updated changelog."
```
mesaugat, line 80:
```suggestion
    printfln "Changelog updated."
```

release.sh
mesaugat: It should be fine.

mesaugat: We lost `labels` from the changelog but it seems to be better in my opinion. Feature not a bug.

release.sh
```diff
@@ -69,6 +83,16 @@ compare_and_release() {
 
   if [ -n "$NEXT" ] && [ "$NEXT" != "false" ]; then
     bump
+    changelog
+
+    setup_git
```
mesaugat, line 88: Why do we need a function? Can we put all the commands inline?

release.sh
```diff
@@ -69,6 +83,16 @@ compare_and_release() {
...
+
+    git add CHANGELOG.md
+    git commit -v --edit -m "${NEXT} Release :tada: :fireworks: :bell:"
+
+    git_push
```
mesaugat, line 93: Same.

samirsilwal:
> @samirsilwal Shouldn't we commit the changelog first, tag the release and then deploy? We are doing deploy first.

um ok. sounds fair will move release.sh run in before deploy

release.sh
samirsilwal: OK resolving

mesaugat:
@samirsilwal Please rebase. The other PR just got merged.

kabirbaidhya (approved):

release.sh
```diff
@@ -68,6 +70,22 @@ compare_and_release() {
   fi
 
   if [ -n "$NEXT" ] && [ "$NEXT" != "false" ]; then
+    changelog
+
+    git config --global user.email "travis@travis-ci.org"
+    git config --global user.name "Travis CI"
+
+    git add CHANGELOG.md
+    git commit -v --edit -m "${NEXT} Release :tada: :fireworks: :bell:" -m "[skip ci]"
```
mesaugat, line 79: @samirsilwal I used this. Much clean, such wow! 🙏 

https://stackoverflow.com/a/5070502/2352587

mesaugat (approved):

PR #105, opened 2021-03-16
Add includeExtension option in migration specific configurations

cham11ng:
**Problem:**

The problem was identified while replacing the knex migrator with sync-db migrator.

Reason: sync-db migrator removes the extension (`.ts, .js`) from filename while inserting in the migrations table but `Knex` migrator by default includes an extension. This corrupts the existing migrations because of the mismatched filename and migrations' table value.

| Knex Migrations Table | sync-db Migrations Table |
| -- | -- |
| 20170918131001_create_employees_table.ts | 20170918131001_create_employees_table |

**Solution:**

For `sourceType` value `javascript` and `typescript`, added `includedExtension` to customize the including/excluding extension.

```diff
  migration:
    directory: 'migration',
    tableName: 'knex_migrations', // Note: This is Knex's default value. Just keeping it same.
    sourceType: 'javascript',
+   includeExtension: true
```

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -27,7 +27,8 @@ export const DEFAULT_CONFIG: Configuration = {
   migration: {
     directory: 'migration',
     tableName: 'knex_migrations', // Note: This is Knex's default value. Just keeping it same.
-    sourceType: 'sql'
+    sourceType: 'sql',
+    includeExtension: false
   }
 };
 
--- a/test/unit/migration/migrator.test.ts
+++ b/test/unit/migration/migrator.test.ts
@@ -221,6 +221,60 @@ describe('MIGRATION: migrator', () => {
       expect(result1[0].methods.down.name).to.deep.equal('down');
     });
 
+    it('should include extension in the resolved object.', async () => {
+      const migrationPath = await mkdtemp();
+
+      const exampleJs = `
+      function up(db) {
+        return db.schema.createTable('demo_table', table => {
+          table.increments('id').primary()
+        });
+      }
+
+      function down(db) {
+        return db.schema.dropTable('demo_table');
+      }
+
+      module.exports = {
+        up: up,
+        down: down
+      }
+      `;
+
+      const exampleTs = `
+      export function up(db: any) {
+        return db.schema.createTable('test_demo', (table: any) => {
+          table.increments('id').primary()
+        });
+      }
+
+      export function down(db: any) {
+        return db.schema.dropTable('test_demo');
+      }
+      `;
+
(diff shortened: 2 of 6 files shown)


test/unit/migration/migrator.test.ts
```diff
@@ -221,6 +221,60 @@ describe('MIGRATION: migrator', () => {
...
+      `;
+
+      // Populate migration files to a directory.
+      await Promise.all([
+        write(path.join(migrationPath, '0001_create_table_users.js'), exampleJs),
+        write(path.join(migrationPath, '0002_alter_table_users_add_gender.ts'), exampleTs),
+        write(path.join(migrationPath, '.gitignore'), '')
+      ]);
+
+      const result = await migratorService.resolveJavaScriptMigrations(migrationPath, 'js', true);
```
mesaugat, line 263: How about we always include the extension for `.js` and `.ts` files and not make it configurable?

mesaugat: @cham11ng @samirsilwal Thoughts?

test/unit/migration/migrator.test.ts
```diff
@@ -221,6 +221,60 @@ describe('MIGRATION: migrator', () => {
       expect(result1[0].methods.down.name).to.deep.equal('down');
     });
 
+    it('should include extension in the resolved object.', async () => {
```
mesaugat, line 224:
```suggestion
    it('should resolve migration entries with correct extension.', async () => {
```

src/constants.ts
```diff
@@ -27,7 +27,8 @@ export const DEFAULT_CONFIG: Configuration = {
   migration: {
     directory: 'migration',
     tableName: 'knex_migrations', // Note: This is Knex's default value. Just keeping it same.
-    sourceType: 'sql'
+    sourceType: 'sql',
+    includeExtension: false
```
mesaugat, line 31: We should remove this configuration and always `includeExtension`. Check comment below.

test/unit/migration/migrator.test.ts
samirsilwal: sounds good. easier shift from knex to sync-db if in use

cham11ng: Totally agreed. I made it configurable to support backward compatibility but to allow migration filenames without extension doesn't make sense. I'll mark this as a bug in migrator and change accordingly.

cham11ng:
I've fixed this issue in another PR.  https://github.com/leapfrogtechnology/sync-db/pull/107

Closing this PR.

PR #107, opened 2021-03-16
Fix migration issue by including file extension

cham11ng:
**Issue:**

The problem was identified while replacing the knex migrator with sync-db migrator.

Reason: sync-db migrator removes the extension (`.ts, .js`) from filename while inserting in the migrations table but `Knex` migrator by default includes an extension. This corrupts the existing migrations because of the mismatched filename and migrations' table value.

| Knex Migrations Table | sync-db Migrations Table |
| -- | -- |
| 20170918131001_create_employees_table.ts | 20170918131001_create_employees_table |

**Solution**

Passing the full filename including an extension in migration object.

### DIFF
--- a/src/migration/service/migrator.ts
+++ b/src/migration/service/migrator.ts
@@ -117,7 +117,7 @@ export async function resolveJavaScriptMigrations(
     const { up, down } = mRequire(path.resolve(migrationPath, filename));
 
     return {
-      name,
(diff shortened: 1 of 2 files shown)


samirsilwal (approved):

mesaugat (approved):
Yay!

PR #117, opened 2021-03-23
Throw an exception to rollback transaction if SQL migration fails

cham11ng:
Problem:

- `db.connection.transaction(async trx => { }` didn't rollback the transaction because no exception was thrown in SQL migration process. In case of migration and rollback `trx.migrate` and `trx.rollback` transaction rollback is handled internally.

- I've to remove the log from `getMigrationName` because it called more than a time which pollutes the console for a large number of migration files.

### DIFF
--- a/examples/node-mssql-programmatic-use/connections.sync-db.json.example
+++ b/examples/node-mssql-programmatic-use/connections.sync-db.json.example
@@ -4,11 +4,11 @@
       "id": "db1",
       "client": "mssql",
       "connection": {
-        "host": "localhost",
+        "host": "db",
         "port": 1433,
         "user": "sa",
         "database": "tempdb",
-        "password": "Password@123"
+        "password": "Secret@123"
       }
     }
   ]
--- a/src/service/execution.ts
+++ b/src/service/execution.ts
@@ -77,7 +77,7 @@ export async function executeOperation<T extends OperationContext>(
   if (result.error) {
     logDb('Result:\n%O', result);
 
-    throw result.error
+    throw result.error;
   }
 
   return result;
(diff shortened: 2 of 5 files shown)


samirsilwal (approved):

aanchal-adk (approved):

src/service/execution.ts
```diff
@@ -74,5 +74,11 @@ export async function executeOperation<T extends OperationContext>(
     await context.params.onFailed(result);
   }
 
+  if (result.error) {
+    logDb('Result:\n%O', result);
+
+    throw result.error;
+  }
+
   return result;
```
mesaugat, line 83: Is this return used anywhere? @cham11ng

cham11ng: @mesaugat I've updated the PR to explicitly commit on success and rollback on failure.

cham11ng: There was a problem while explicitly committing in mssql may be it something in `tedious` but it worked in Postgres.
Reverted chages.

cham11ng: > Is this return used anywhere?

On this part, I've updated the PR again and tested it locally.

samirsilwal (approved):

examples/node-mssql-programmatic-use/connections.sync-db.json.example
```diff
@@ -4,11 +4,11 @@
       "id": "db1",
       "client": "mssql",
       "connection": {
-        "host": "localhost",
+        "host": "db",
```
mesaugat, line 7: @samirsilwal sneaking his stuff into another PR. LOL!

src/util/promise.ts
```diff
@@ -25,7 +25,11 @@ export async function runSequentially<T>(promisers: Promiser<T>[]): Promise<T[]>
 
       result.push(value);
     } catch (err) {
-      throw err;
+      if (!err.result) {
```
mesaugat, line 28: Still gimmicky. Does the job for now though.

mesaugat (approved):

PR #119, opened 2021-03-26
Add ability to provide custom configuration source

samirsilwal:
## Changes

- Added `config-source` command that which argument act as a new configuration source for successive sync-db commands
**yarn sync-db config-source sync-db-test**
- The feature simple create a copy of the config file & reads configuration from it. 
- To see the current configuration source , we can  use 
**yarn sync-db config-source**
- To reset the configuration cache we can use 
- **yarn sync-db config-source --reset**

Screenshot from 2021-03-26 08-50-42
Screenshot from 2021-03-26 08-51-43
Screenshot from 2021-03-26 08-52-08
Screenshot from 2021-03-26 08-52-51


cham11ng (commented):
@samirsilwal
Using default `sync-db.yml` makes sense that's what sync-db supports but updating the default configuration file might be magic to the user sometimes. You are making sync-db commands stateful might be risky to the user, instead, I prefer explicitly passing the configuration file path something like `--config=`. Any thoughts? cc @mesaugat @aanchal-adk @kabirbaidhya

mesaugat:
I've not looked at the PR but have read the description and checked the screenshots. We should explicitly be able to take the configuration as @cham11ng said. 

So the implementation should be. sync-db shouldn't do the work of remembering the configuration that the user set at some point of time. 

1. The default `sync-db.yml` should always work. It should check the working directory for the configuration file.
2. Explicitly we should be able to pass a custom configuration using `--config=sync-db-for-something.yml` or `-c` (Let's make sure all sync-db configurations are prefixed with `sync-db-*`

@samirsilwal @cham11ng

samirsilwal:
closing

PR #187, opened 2022-12-23
Fail synchronization if any error occurs while running sql scripts

thesamiroli:
**Problem**
While syncing database objects like views, and functions with sync-db, if there were any problems with the query files, then sync-db would not throw any errors but rather complete successfully and the database objects would not be created. That would lead us to believe that the operation was successful because the problems won't be revealed to us.

**Change**
This change fixes the problem by throwing errors if there were any errors in the synchronization scripts.

**Before, even in the case of problems**
image

**After**
image

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -119,7 +119,7 @@ export async function resolveConnections(config: Configuration, resolver?: strin
   } else if (resolver || config.connectionResolver) {
     connections = await resolveConnectionsUsingResolver(resolver || config.connectionResolver, config);
   } else {
-    log('Connections file not provided. Trying to load connection details from env.');
+    log('Connections file not provided. Loading connection details from env.');
 
     connections = resolveConnectionsFromEnv();
   }
--- a/src/config.ts
+++ b/src/config.ts
@@ -119,7 +119,7 @@ export async function resolveConnections(config: Configuration, resolver?: strin
   } else if (resolver || config.connectionResolver) {
     connections = await resolveConnectionsUsingResolver(resolver || config.connectionResolver, config);
   } else {
-    log('Connections file not provided.');
+    log('Connections file not provided. Trying to load connection details from env.');
 
     connections = resolveConnectionsFromEnv();
   }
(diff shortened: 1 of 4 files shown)


src/service/sync.ts
```diff
@@ -41,7 +41,7 @@ async function setup(trx: Knex.Transaction, context: SynchronizeContext): Promis
     log('PRE-SYNC: End');
   }
 
-  // Run the migration scripts.
+  // Run the synchronization scripts.
```
govinda744, line 44:
```suggestion
  // Run the sync scripts for views, functions, and xwalks.
```

sanishmaharjan (approved):

src/util/promise.ts
```diff
@@ -27,13 +29,16 @@ export async function runSequentially<T>(promisers: Promiser<T>[]): Promise<T[]>
...
       if (errors.originalError) {
         result.push(errors.originalError);
-      } else if (errors.error) {
-        result.push(errors.error);
-      } else {
-        throw err;
       }
+
+      result.push(errors);
+      throw flatten(result.filter(r => !isNil(r)));
```
govinda744, line 41: Maybe separate out the results and errors array here. Making the result and errors array the same is confusing and not readable.

sanishmaharjan:
@thesamiroli Build Failed.

src/service/sync.ts
samirsilwal: Don't think we need to all object names in comment. sync-db support triggers sync as well recently. so a generic comment should be enough IMO ` Run the synchronization scripts.`

samirsilwal:
> @thesamiroli Build Failed.

Looks like test needs to be updated

src/config.ts
```diff
@@ -119,7 +119,7 @@ export async function resolveConnections(config: Configuration, resolver?: strin
   } else if (resolver || config.connectionResolver) {
     connections = await resolveConnectionsUsingResolver(resolver || config.connectionResolver, config);
   } else {
-    log('Connections file not provided.');
+    log('Connections file not provided. Trying to load connection details from env.');
```
samirsilwal, line 122:
```suggestion
    log('Connections file not provided. Loading connection details from env.');
```

Kushalkhadka7 (approved):

sanishmaharjan (approved):

mesaugat (approved):
Looks good.

PR #188, opened 2022-12-27
Skip synchronization of a database object if .dropped is appended on its name

thesamiroli:
**Problem**
The only way we have right now to remove an already synced database is to first remove the view manually, and then remove it from the config file and remove the actual file. 

**Change**
With this change, we can just append `.dropped` to any database object in the config file, and then sync-db will only drop the view/function/proc and not synchronize (up) it. We can remove that object from the config file immediately after synchronization or after some point in time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,8 @@ _See code: [src/commands/prune.ts](https://github.com/leapfrogtechnology/sync-db
 
 ## `sync-db synchronize`
 
-Synchronize all the configured database connections.
+Synchronize all the configured database connections. If `.dropped` is appended with an database object's name,
+then that object will only be dropped, and not synchronized.
 
 ```
 USAGE
@@ -279,7 +280,7 @@ basePath: /path/to/sql
 
 sql:
   - schema/<schema_name>.sql
-  - function/<schema_name>/<function_name>.sql
+  - function/<schema_name>/<function_name>.sql.dropped #While syncing this will only be dropped, not upped.
   - procedure/<schema_name>/<procedure_name>.sql
 ```
 
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@leapfrogtechnology/sync-db",
   "description": "Command line utility to synchronize and version control relational database objects across databases",
-  "version": "1.2.0",
+  "version": "1.1.1",
   "license": "MIT",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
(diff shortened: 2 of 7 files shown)


sanishmaharjan (approved):

src/service/sqlRunner.ts
```diff
@@ -79,7 +89,10 @@ export function extractSqlFileInfo(filePath: string): SqlFileInfo {
   const fileParts = filePath.split('/');
   const fileName = fileParts.pop() || '';
   const [type, schema] = fileParts;
-  const name = fileName.replace('.sql', '');
+
+  // File name might contain `.dropped` at the end which we need to remove in case of dropping the database objects.
+  let name = fileName.replace(DROP_ONLY_OBJECT_TERMINATOR, '');
+  name = name.replace('.sql', '');
```
samirsilwal, line 95:
```suggestion
  const name = fileName.replace(/(.sql)|(.dropped)/g, '');
```

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@leapfrogtechnology/sync-db",
   "description": "Command line utility to synchronize and version control relational database objects across databases",
-  "version": "1.1.1",
+  "version": "1.1.2",
```
mesaugat, line 4: This change includes a new feature. Can you update the minor version instead? @thesamiroli 

```suggestion
  "version": "1.2.0",
```

samirsilwal:
@thesamiroli Lint failed

samirsilwal (approved):

package.json
thesamiroli: Okay!

README.md
```diff
@@ -279,7 +280,7 @@ basePath: /path/to/sql
 
 sql:
   - schema/<schema_name>.sql
-  - function/<schema_name>/<function_name>.sql
+  - function/<schema_name>/<function_name>.sql.dropped #While syncing this will only be dropped, not upped.
```
mesaugat, line 283: - function/<schema_name>/<function_name>.sql.dropped # While synchronizing this object will only be dropped, not created

README.md
```diff
@@ -200,7 +200,8 @@ _See code: [src/commands/prune.ts](https://github.com/leapfrogtechnology/sync-db
 
 ## `sync-db synchronize`
 
-Synchronize all the configured database connections.
+Synchronize all the configured database connections. If `.dropped` is appended with an database object's name,
```
mesaugat, line 203: @thesamiroli Can we just name it `.drop` instead of `.drop` because the sync-db synchronize will always try to drop the object even if it is not available.

PR #198, opened 2023-04-27
Upgrade `knex(v2.4.2)` and `typescript(v5.0.4)` to latest version

kshitish182:
Upgrade `knex(v2.4.2)` and `typescript(v5.0.4)` to latest version

Stuffs done

* Bumped knex to v2.4.2
* Bumped typescript to v5.0.4
* Updated `knex` imports
* Fixed some type issues
* Updated connection config keys for `mssql` based on the respective update on `knex`

I have tested all the major features on the example apps and everything looks good. Ran the tests locally as well and all passed.

mssql-ts-1
mssql-ts-2
mssql-1
mssql-2


.gitignore
```diff
@@ -7,3 +7,6 @@
 /tmp
 node_modules
 /coverage
+
+# IDE settings
+/.vscode
```
mesaugat, line 12:
```suggestion
.vscode
```

examples/node-app-mssql-ts/package.json
```diff
@@ -12,12 +12,12 @@
     "rollback": "../../bin/run-dev.sh migrate-rollback"
   },
   "dependencies": {
-    "@leapfrogtechnology/sync-db": "^1.0.0-beta.10",
-    "knex": "^0.95.2",
+    "@leapfrogtechnology/sync-db": "1.2.1",
```
mesaugat, line 15: This should change to `2.0.0` after we merge this change? So this means `sync-db` works without any changes and just upgrading the knex peer dependency in the example file?

```suggestion
    "@leapfrogtechnology/sync-db": "1.2.1",
```

examples/node-mssql-programmatic-use/src/sync-db.js
```diff
@@ -1,4 +1,4 @@
-import { synchronize, loadConfig, resolveConnections } from '@leapfrogtechnology/sync-db';
+const { synchronize, loadConfig, resolveConnections } = require('@leapfrogtechnology/sync-db');
```
mesaugat, line 1: Why require? Can we not use imports?

src/config.ts
```diff
@@ -58,7 +58,7 @@ export async function loadConfig(configFilename: string = CONFIG_FILENAME): Prom
 
   log('Resolved config file.');
 
-  const loaded = mergeDeepRight(DEFAULT_CONFIG, loadedConfig) as Configuration;
+  const loaded = mergeDeepRight<Configuration, Configuration>(DEFAULT_CONFIG, loadedConfig) as unknown as Configuration;
```
mesaugat, line 61: as unknown as Configuration? Can we not only use the generics to type this correctly?

src/config.ts
```diff
@@ -131,6 +131,7 @@ export async function resolveConnections(config: Configuration, resolver?: strin
       client,
       connection: {
         host: (connection as any).host,
+        server: (connection as any).server,
```
mesaugat, line 134: Maybe we need to type things better. Let's try to avoid `connection as any` in here.

examples/node-app-mssql-ts/package.json
kshitish182: I used the build zip created by `yarn pack` to test my changes. I have reverted that here and instead upgraded it to the latest version. Yes, will need to update this to v2.0.0 after the package is released

examples/node-mssql-programmatic-use/src/sync-db.js
kshitish182: did it to test the programmatic `synchronize()` using node. So, changed it to require

src/config.ts
```diff
@@ -214,7 +215,10 @@ export function resolveConnectionsFromEnv(): ConnectionConfig[] {
       database: process.env.DB_NAME,
       options: {
         encrypt: process.env.DB_ENCRYPTION === 'true'
-      }
+      },
+      // Knex config for mssql uses these keys in place of "host" and "user"
+      server: process.env.DB_SERVER,
```
mesaugat, line 220: Let's still use DB_HOST as the environment variable to avoid confusion? Also, does server and userName keys work for postgres and mssql.

```suggestion
      server: process.env.DB_HOST,
```

src/util/db.ts
```diff
@@ -48,11 +48,15 @@ export function createInstance(config: ConnectionConfig): Knex {
...
 
-  const { host, database } = config.connection as any;
+  const { host, server, database, userName } = config.connection as any;
 
-  log(`Connecting to database: ${host}/${database}`);
+  log(`Connecting to database: ${server ? server : host}/${database}`);
 
-  return Knex({
+  if (config.client === 'mssql' && (!server || !userName)) {
+    throw new Error(`Db client ${config.client} requires key "server" and "userName".`)
```
mesaugat, line 56: Hmm so this is only for mssql. What a mess.

src/config.ts
kshitish182: Yes, I thought so too, But mergeDeepRight returns some other type. `MergeDeep<Configuration, Configuration>`, so had to type cast it here.

mesaugat:
@kshitish182 One suggestion. Can we separate the core changes from the example changes. Let's create another PR just for the example changes.

src/config.ts
kshitish182: Yes, that's the confusion I had. Should we handle this change in the library or let the user handle it? No, postgres still requires `host` and 'user`. This is specifically for mssql. Don't know why Knex did this. 🤷

src/util/db.ts
kshitish182: Exactly

mesaugat:
@kshitish182 I still see changes in the example directory. Can you take a look?

src/config.ts
mesaugat: Let's still use DB_HOST as our variable. This is under our control and is less confusing.

src/service/configInjection.ts
```diff
@@ -74,6 +74,12 @@ export async function setup(trx: Knex.Transaction, context: SynchronizeContext):
 
   const values = convertToKeyValuePairs(injectedConfig.vars);
 
+  if (!values.length) {
+    log(`Config not available. Skipping insertion on ${INJECTED_CONFIG_TABLE} table.`);
+
+    return;
```
mesaugat, line 80: Let's exit here with `process.exit(1)` because the setup function's return type is void.

src/util/db.ts
```diff
@@ -48,11 +48,15 @@ export function createInstance(config: ConnectionConfig): Knex {
     throw new Error('The provided connection already contains a connection instance.');
   }
 
-  const { host, database } = config.connection as any;
+  const { host, server, database, userName } = config.connection as any;
```
mesaugat, line 51: @kshitish182 This is slightly irritating. Can you check why config.connection has to be casted to any? We should avoid any.

src/config.ts
```diff
@@ -58,7 +58,10 @@ export async function loadConfig(configFilename: string = CONFIG_FILENAME): Prom
 
   log('Resolved config file.');
 
-  const loaded = mergeDeepRight(DEFAULT_CONFIG, loadedConfig) as Configuration;
+  const loaded = (mergeDeepRight<Configuration, Configuration>(
+    DEFAULT_CONFIG,
+    loadedConfig
+  ) as unknown) as Configuration;
```
mesaugat, line 64: .

src/util/db.ts
```diff
@@ -48,11 +48,15 @@ export function createInstance(config: ConnectionConfig): Knex {
...
 
-  log(`Connecting to database: ${host}/${database}`);
+  log(`Connecting to database: ${server ? server : host}/${database}`);
 
-  return Knex({
+  if (config.client === 'mssql' && (!server || !userName)) {
+    throw new Error(`Db client ${config.client} requires key "server" and "userName".`);
```
mesaugat, line 56: Hmm I think we should simplify the API by still using `user` and `host` but just internally use `host` and `userName`. Meaning, we should let sync-db users still pass `host` and `user` and we internally map it to `server` and `userName` when passing it to knex?

What do you think? @kshitish182

src/migration/KnexMigrationSource.ts
```diff
@@ -47,15 +49,16 @@ class KnexMigrationSource {
...
    */
-  getMigration(name: string) {
+  getMigration(name: string): Promise<MigrationRunner> {
     const migration = this.migrationContext.get(name);
 
     this.log(`getMigration - ${name}`);
     this.log(`getMigration - resolve: %o`, migration);
 
-    return migration;
+    return Promise.resolve(migration);
```
mesaugat, line 61: Do we need to return a Promise? knex documentation just returns the migration context here. 

https://knexjs.org/guide/migrations.html#custom-migration-sources

test/unit/migration/KnexMigrationSource.test.ts
```diff
@@ -54,7 +54,7 @@ describe('MIGRATION: KnexMigrationSource', () => {
       const instance = await getInstance();
       const dbStub = {} as any;
 
-      const migration = instance.getMigration('mgr_002');
+      const migration = await instance.getMigration('mgr_002');
```
mesaugat, line 57: Hmm ... there is something off here. Might not need `getMigration` to return a `Promise`.

src/migration/KnexMigrationSource.ts
kshitish182: Yes, this part bothered me as well. The type definition says it returns a Promise though. 🤯 . 
migrationtype

src/util/db.ts
kshitish182: Yes, will look into it. We probably knex's type definition for this.

test/unit/migration/KnexMigrationSource.test.ts
kshitish182: `getMigration()` type definition says it returns a promise 😅

src/util/db.ts
kshitish182: Yes, I think we can handle it this way. Will make the changes here.

src/service/configInjection.ts
kshitish182: Won't this abruptly end the process? We need to just skip this step and let the task continue.

test/unit/migration/KnexMigrationSource.test.ts
mesaugat: Understood. The type definition is a little off than the documentation. Cool!

src/migration/KnexMigrationSource.ts
mesaugat: Understood.

src/service/configInjection.ts
mesaugat: Ah good point. Let's keep it as it is.

src/config.ts
mesaugat: If there's a problem here, we should use `lodash.merge` instead of Ramda. We don't need the whole ramda thing for such a tiny functionality.

kshitish182: Updating ramda to `v0.29.0` solved the type issue. Bumping the ramda version

src/config.ts
```diff
@@ -130,8 +131,8 @@ export async function resolveConnections(config: Configuration, resolver?: strin
       id,
       client,
       connection: {
-        host: (connection as any).host,
-        database: (connection as any).database
+        host: (connection as Knex.ConnectionConfig).host,
```
mesaugat, line 134: Still doesn't cut it in my opinion. Why doesn't connection have a concrete type?

mesaugat: What is the type for `connections` can you check?

mesaugat: We shouldn't be doing `as` as much as possible.

kshitish182: The type for connection is `string | StaticConnectionConfig | ConnectionConfigProvider`. The string for the connection string, `StaticConnectionConfig` is the type for various db connection config ORed together and `ConnectionConfigProvider` basically returns `StaticConnectionConfig` using both sync and async function.

The problem here is with these multiple types for the single key, we have to force the compiler to take in the type we are using.

mesaugat (approved):

PR #200, opened 2023-05-01
Bump knex,typescript and sync-db version in all example apps

kshitish182:
Bump knex and typescript version in all example apps

Stuffs done:

* Bumped `knex` to `v2.4.2`
* Bumped `typescript` to `v5.0.4` for apps using typescript
* Bumped `sync-db` to `v2.0.0`
* Replaced `mssql` by the latest version of `tedious`

### DIFF
--- a/examples/node-app-mssql-ts/Dockerfile
+++ b/examples/node-app-mssql-ts/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:14-slim
+FROM node:16-slim
 
 WORKDIR /app/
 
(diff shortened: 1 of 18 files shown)


examples/node-app-mssql-ts/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:14-slim
+FROM node:16-slim
```
mesaugat, line 1: Node 16 is going away soon. Let's use Node 18.

```suggestion
FROM node:18-slim
```

samirsilwal (approved):

PR #203, opened 2023-05-17
Convert travis file to actions

### LINKED ISSUE #139: Move from Travis CI to GitHub Actions
No one is willing to wait 182 days for a CI job to finish.

sbimochan:
resolves #139 

- Please review carefully on `echo` parts if it would produce the same behavior
- The secrets needs to be defined in github

> wanted to make changes on travis.yml file itself so that diff could be viewed side by side but would create extra PRs and gave syntax error in .travis, later thought it would be a bad file history haha.


sbimochan:
@mesaugat Travis CI is set to required

.github/workflows/master.yml
```diff
@@ -0,0 +1,87 @@
...
+        run: yarn test:coverage
+
+      - name: Upload coverage to Codecov
+        run: yarn codecov
+
+      - name: Determine branch
+        id: branch
+        run: echo "::set-output name=branch::$(if [ ${{ github.event_name }} == 'pull_request' ]; then echo ${{ github.event.pull_request.head.ref }}; else echo ${{ github.ref }}; fi)"
+
+  deploy:
```
mesaugat, line 65: @sbimochan Rename this to publish. We need to publish the package which is a better name.

.github/workflows/master.yml
```diff
@@ -0,0 +1,87 @@
...
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [16, 18]
```
mesaugat, line 17:
```suggestion
        node-version: [16, 18, 20]
```

.github/workflows/master.yml
```diff
@@ -0,0 +1,87 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: 'npm'
```
mesaugat, line 27: We use yarn to install packages.

```suggestion
          cache: 'yarn'
```

.github/workflows/master.yml
```diff
@@ -0,0 +1,87 @@
...
+      - name: Install github-changelog-generator
+        run: gem install github_changelog_generator --user-install
+
+      - name: Install dependencies
+        run: yarn
+
+      - name: Set Ribby as git user
+        run: |
+          git config user.name "Ribby"
+          git config user.email "bot@lftechnology.com"
```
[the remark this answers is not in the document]
mesaugat: Can you remove the `git config` step inside of `release.sh`. It is being done there too.

.github/workflows/master.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      - name: Test coverage
+        run: yarn test:coverage
+
+      - name: Upload coverage to Codecov
+        run: yarn codecov
+
+      - name: Determine branch
+        id: branch
+        run: echo "::set-output name=branch::$(if [ ${{ github.event_name }} == 'pull_request' ]; then echo ${{ github.event.pull_request.head.ref }}; else echo ${{ github.ref }}; fi)"
```
mesaugat, line 63: Why do we need this?

.github/workflows/master.yml
```diff
@@ -0,0 +1,87 @@
+name: Build and Release from Master
```
mesaugat, line 1:
```suggestion
name: Build and Publish
```

mesaugat: Can we also rename the file to `build-publish.yml`?

.github/workflows/master.yml
samirsilwal, line 63: Lets remove use of set-output its deprecated  https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

.github/workflows/master.yml
```diff
@@ -0,0 +1,87 @@
...
+        uses: actions/checkout@v3
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: node
+          cache: 'npm'
+
+      - name: Run release script
+        run: ../../release.sh compare_and_release
```
samirsilwal, line 82: This path is not correct probably. Checkout action should set current directory to that of repo home so ./release.sh should be enough

.github/workflows/master.yml
```diff
@@ -0,0 +1,87 @@
...
+        run: |
+          sudo apt update
+          sudo apt install hub
+          hub --version
+
+      - name: Install github-changelog-generator
+        run: gem install github_changelog_generator --user-install
+
+      - name: Install dependencies
+        run: yarn
```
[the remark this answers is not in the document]
mesaugat: We should cache yarn instead as we have a yarn.lock file.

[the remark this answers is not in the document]
samirsilwal: true 😅

.github/workflows/master.yml
```diff
@@ -0,0 +1,87 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: node
+          cache: 'npm'
```
samirsilwal, line 79: .

.github/workflows/master.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      - name: Run release script
+        run: ../../release.sh compare_and_release
+
+      - name: Deploy to NPM
+        run: |
+          echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > ~/.npmrc
+          npm publish
```
samirsilwal, line 87: yarn for publish as well

mesaugat:
@sbimochan Can you use RIBBY_GITHUB_TOKEN as the environment variable instead of GITHUB_TOKEN inside of `release.sh` file.

I've already included the secrets in GitHub secrets.

.github/workflows/build-publish.yml
```diff
@@ -0,0 +1,84 @@
...
+          node-version: ${{ matrix.node-version }}
+          cache: 'yarn'
+
+      - name: Install hub
+        run: |
+          sudo apt update
+          sudo apt install hub
+          hub --version
+
+      - name: Install github-changelog-generator
```
mesaugat, line 35:
```suggestion
      - name: Install GitHub changelog generator
```

.github/workflows/build-publish.yml
```diff
@@ -0,0 +1,84 @@
...
+        run: gem install github_changelog_generator --user-install
```
mesaugat, line 36: Don't think `gem` would be available. Please check.

.github/workflows/build-publish.yml
```diff
@@ -0,0 +1,84 @@
...
+        run: yarn build
+
+      - name: Test coverage
+        run: yarn test:coverage
+
+      - name: Upload coverage to Codecov
+        run: yarn codecov
+
+  publish:
+    name: Publish this as a package
```
mesaugat, line 62:
```suggestion
    name: Publish @leapfrogtechnology/sync-db package
```

mesaugat: Or simply "Publish package"

.github/workflows/build-publish.yml
mesaugat: Also, this is only required in the publish step.

sbimochan: added a step before

mesaugat (approved):

mesaugat:
@sbimochan Looking good. Will merge, I think there are few places where it still might not work during publish but that we can fix later.

Thanks for sending the PR. We finally don't have to wait 3 years for Travis CI to finish.

Issue #204, opened 2023-06-02
[object object] seen while getting migration lists


samirsilwal:
## Issue
- Migrations not logged properly

## Replication

- Run `migrate-list` command for given set of migration

![image](https://github.com/leapfrogtechnology/sync-db/assets/35683880/3f8e48cd-379d-4395-9d32-6f9b941a2edf)

samirsilwal:
Fixed by https://github.com/leapfrogtechnology/sync-db/pull/206

PR #206, opened 2023-06-07
Fix the type displayed instead of migration name with migrate-list

samirsilwal:
## Changes
image

- The result of running the migration list returns data as above now. Previously, it used to be just the array of names
- I am not sure about the cause, but don't see any related changes in recent commits

## After fix

image

Fixes https://github.com/leapfrogtechnology/sync-db/issues/204

### DIFF
--- a/src/commands/migrate-list.ts
+++ b/src/commands/migrate-list.ts
@@ -36,7 +36,9 @@ class MigrateList extends Command {
 
     // Completed migrations.
     for (const item of completedList) {
-      await printLine(cyan(`   • ${item}`));
+      const completedMigrationName = typeof item === 'string' || item instanceof String ? item : item?.name;
+
+      await printLine(cyan(`   • ${completedMigrationName}`));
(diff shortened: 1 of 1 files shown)


mesaugat:
@samirsilwal If this works, we can change the base branch to `fix/remove-error-log-when-dir-does-not-exist`. Trying to make a 2.0.1 release there.

samirsilwal:
> @samirsilwal If this works, we can change the base branch to `fix/remove-error-log-when-dir-does-not-exist`. Trying to make a 2.0.1 release there.

Yeah it works, i will finalize the PR

mesaugat (approved):
Looks good to me.

mesaugat:
@samirsilwal Should we do the same for remaining migration list?

samirsilwal:
> @samirsilwal Should we do the same for remaining migration list?

The remaining are array of string so it works with current code

PR #229, opened 2024-04-27
Update --only flag to support multiple connection filter

samirsilwal:
## Context

- For multiple connection provided to sync-db, it was difficult to filter a subset of connection for which we require to synchronize the migrations. `--only` flag supported only a single connectionId which is a bit hectic

## Changes

- Update `--only` to support multiple `,` separated connection identifiers
- Any unwanted connection ids also get filtered if provided

### connections.sync-db.json
```
{
  "connections": [
    {
      "id": "db1",
      "client": "mssql",
      "connection": {
        "host": "mssql",
        "port": 1433,
        "user": "sa",
        "database": "tempdb",
        "password": "Secret@123"
      }
    },
    {
      "id": "db2",
      "client": "mssql",
      "connection": {
        "host": "mssql",
        "port": 1433,
        "user": "sa",
        "database": "tempdb1",
        "password": "Secret@123"
      }
    },
    {
      "id": "db3",
      "client": "mssql",
      "connection": {
        "host": "mssql",
        "port": 1433,
        "user": "sa",
        "database": "tempdb2",
        "password": "Secret@123"
      }
    },
    {
      "id": "db4",
      "client": "mssql",
      "connection": {
        "host": "mssql",
        "port": 1433,
        "user": "sa",
        "database": "tempdb3",
        "password": "Secret@123"
      }
    }
  ]
}

```

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@leapfrogtechnology/sync-db",
   "description": "Command line utility to synchronize and version control relational database objects across databases",
-  "version": "2.0.1",
+  "version": "2.0.2",
   "license": "MIT",
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
--- a/src/api.ts
+++ b/src/api.ts
@@ -247,7 +247,7 @@ function filterConnectionsAsRequired(
   connections: ConnectionReference[],
   filterConnectionIds?: string
 ): ConnectionReference[] {
-  const trimmedFilterConnectionIds = filterConnectionIds?.split(',').map(id => id.trim());
+  const trimmedFilterConnectionIds = Array.from(new Set(filterConnectionIds?.split(',').map(id => id.trim())));
   const formattedFilterConnectionIds = trimmedFilterConnectionIds?.join(', ');
 
   log(`Filter(s) (--only=) ${formattedFilterConnectionIds}`);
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@leapfrogtechnology/sync-db",
   "description": "Command line utility to synchronize and version control relational database objects across databases",
-  "version": "2.0.1",
+  "version": "2.0.2",
```
sbimochan, line 4:
```suggestion
  "version": "2.1.0",
```

sbimochan (approved):

src/api.ts
```diff
@@ -240,31 +240,38 @@ export async function migrateList(
...
   }
 
-  const filteredList = connections.filter(connection => connection.id === filterConnectionId);
+  const filteredList = connections.filter(connection => trimmedFilterConnectionIds?.includes(connection.id));
 
   if (filteredList.length === 0) {
     const available = connections.map(({ id }) => id);
 
-    throw new Error(`No connections found for given id "${filterConnectionId}. Available ids are: ${available}`);
+    throw new Error(
```
silwalanish, line 267: Do we need to handle case if at least one connection id is found and some of the ids aren't found? 

Example: For --only=id1,id2,id3, if id1 and id3 is present and id2 isn't present. Should we log some error for id3 or stop the process altogether? 

Also what if one does --only=id1,id1. Would it run sync-db twice for same connection. Let's remove duplicates just in case.

samirsilwal: Yes unique id would be good. Also will add a log in case we add entry for invalid connection id

src/api.ts
```diff
@@ -240,31 +240,42 @@ export async function migrateList(
  * Check the filter condition and apply filter if required.
  *
  * @param {ConnectionReference[]} connections
- * @param {string} [filterConnectionId]
+ * @param {string} [filterConnectionIds]
```
mesaugat, line 243: Let's rename this param to `connectionIds` so that it is more readable. The function name already says `filterConnectionsAsRequired`?

```suggestion
 * @param {string} [connectionIds]
```

src/api.ts
```diff
@@ -240,31 +240,42 @@ export async function migrateList(
...
  * @returns {ConnectionReference[]}
  */
 function filterConnectionsAsRequired(
```
mesaugat, line 246:
```suggestion
function filterConnections(
```

mesaugat (approved):

PR #233, opened 2024-07-17
Introduce new command to run manual scripts

samirsilwal:
## Changes

- Introduce new command `run-script`
- Accepts `--file` as flag
- Can be used to run manual scripts in JS/TS or SQL
- For SQl, simply run the script present in the given file
- For JS/TS, a function `main` is expected to be exported which gets knex connection and identifier as args. example
```
export async function main(conn: Knex.Transaction, id: string) {
  await conn.......

  log(`Completed script for ${id}`)
}
```
- Other existing flags like `--only` -> to run script for specific connection and `--dry-run` -> to dry run the script works as usual
- Update unit tests

## Todo
- Update readme
- Add demo ss

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -6,12 +6,15 @@
  * are also meant to be the public interface for the developers using it as a package.
  */
 
+import { Knex } from 'knex';
+
 import * as init from './init';
 import { log } from './util/logger';
 import { existsDir } from './util/fs';
 import { withTransaction, mapToConnectionReferences, DatabaseConnections } from './util/db';
 
 import Configuration from './domain/Configuration';
+import { RunScriptParams } from './domain/RunScriptParams';
 import SynchronizeParams from './domain/SynchronizeParams';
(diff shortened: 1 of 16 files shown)


src/api.ts
```diff
@@ -6,21 +6,75 @@
...
+ * Run scripts for all the configured database connections.
+ *
+ * @param {Configuration} config
+ * @param {DatabaseConnections} conn
+ * @param {RunScriptParams} options
+ * @returns {Promise<OperationResult[]>}
+ */
+export async function runScriptAPI(config: Configuration, conn: DatabaseConnections, options?: RunScriptParams) {
+  log('Run Script');
+  const scriptPath = getManualScriptPath(config);
```
SafalPandey, line 44: Let's not name things as manual here. Just *getScriptPath* should be good.

mesaugat:
Plan to revisit later. Don't need this feature at the moment.

PR #237, opened 2024-10-11
Use relative base path respective to config file

samirsilwal:
closes #225

## Changes

- Update base path to be relative to sync-db config file location
- Update name extraction condition
- Update default value for base path

## Test
- directory structure
image

1. CMD yarn sync-db migrate-list --config=./src/sync-db.yml
image

2. CMD yarn sync-db migrate-rollback --config=./src/sync-db.yml
image

3. CMD yarn sync-db migrate-latest --config=$(pwd)/src/sync-db.yml
image

4. CMD yarn sync-db synchronize --config=$(pwd)/src/sync-db.yml
image

5. CMD yarn sync-db make test_update_example --config=$(pwd)/src/sync-db.yml
image

- directory structure 2
image

1. CMD yarn sync-db migrate-list --config=./sync-db.yml
image

2. CMD yarn sync-db migrate-rollback --config=./sync-db.yml
image

3. CMD yarn sync-db migrate-latest --config=$(pwd)/sync-db.yml
image

4. CMD yarn sync-db synchronize --config=$(pwd)/sync-db.yml
image

PR #238, opened 2024-10-15
Update build workflow to include codecov upload step

samirsilwal:
## Changes

- Update `build-publish` workflow to use codecov upload action instead of codecov package
- Remove deprecated `codecov` package
- New version: `2.2.0` to include changes in #237

### DIFF
--- a/.github/workflows/build-publish.yml
+++ b/.github/workflows/build-publish.yml
@@ -47,7 +47,13 @@ jobs:
         run: yarn test:coverage
 
       - name: Upload coverage to Codecov
-        run: yarn codecov
+        uses: codecov/codecov-action@v4
+        with:
+          fail_ci_if_error: true 
+          directory: ./coverage
+          files: ./lcov.info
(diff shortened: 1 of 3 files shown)


samirsilwal:
@mesaugat Need to add token to secrets with name CODECOV_TOKEN

mesaugat:
@samirsilwal Thanks. Let me look at that. Can you also update the package.json version to 2.2.0 and commit the change. The workflow only publishes a new package version when package.json version is a new one.

samirsilwal:
> @samirsilwal Thanks. Let me look at that. Can you also update the `package.json` version to `2.2.0` and commit the change. The workflow only publishes a new package version when package.json version is a new one.

Sure will make the changes

mesaugat (approved):
@samirsilwal Codecov now works. I added the token and integrated the Codecov app in the repository.

Also, changed the package version to 2.2.0 to publish a new package version.

Issue #242, opened 2024-10-22
When incorrect sub command is given, the script runs forever


samirsilwal:
## Context

- When we provide any invalid sub command to sync-db we run into a infinite loop until the command timeout

### Example
![image](https://github.com/user-attachments/assets/9282774d-505c-4316-8c3b-9cada88614da)

### Steps to replicate

Use any not existing command with sync-db

```
yarn sync-db migrate-gg
```

KaushalGoud:
Hi @mesaugat  — following up on this. I tested reproducing the original bug using DEBUG=* sync-db gg (matching @samirsilwal 's repro from the now-closed #243) on the current master. It no longer hangs — it fails immediately with command gg not found.

Looking at the debug output, a command_not_found hook is now firing correctly in @oclif/config, so it looks like the oclif upgrade you mentioned wanting already happened via dependency bumps, and this issue's underlying bug is resolved.

Tested on Node v22 and v26, current package.json dependency versions. Happy to close this if you agree — or I can add a regression test to lock in the behavior so it doesn't silently break again.

KaushalGoud:
Hello Sir, Can you also guide me in the interview process of leapfrog. Like
the DSA , Development which is more valuable to the company.

PR #243, opened 2024-10-26
Handle the invalid command resulting in infinite loop

samirsilwal:
## Changes

- Check if the command provided to sync-db in valid one 
- Show the available command in case of invalid entry

- Fixes the infinite loop while unknown command is given to sync-db

### Before

`DEBUG=* yarn sync-db gg`
image

### After

`DEBUG=* yarn sync-db gg`

image


samirsilwal:
Resolves https://github.com/leapfrogtechnology/sync-db/issues/242

samirsilwal:
Will add the tests

mesaugat:
Should move to newer oclif version instead and add this in if it doesn't exist.

PR #259, opened 2026-01-20
Add command log feature in sync db

samirsilwal:
## Changes

- Add logs of command run in sync-db
- Update commands using db connection to store the status of command run in a connection specific log table

image


Kushalkhadka7 (approved):

Kushalkhadka7:
@samirsilwal Can we add description on the js docs.

src/service/runLogger.ts
```diff
@@ -0,0 +1,124 @@
...
+  run_id: string;
+  run_date: Date;
+  error?: string;
+  connection_id?: string;
+  is_successful: boolean;
+  command_type: CommandType;
+  metadata?: Record<string, any>;
+}
+
+const TABLE_NAME = '__sync_db_run_logs';
```
mesaugat, line 30:
```suggestion
const TABLE_NAME = '__sync_db_logs';
```

mesaugat:
Seems like this is not necessary right now.

PR #261, opened 2026-01-21
feat: add --sync-files flag while synchronzing

samirsilwal:
## Changes

- Add --sync-files flag to synchronize command
- It is list of objects that is to be synchronized 
- This introduces an explicit way of providing only the specific object to be synced
- The path should be list of relative paths to base path given in sync-db yaml
- When there is non empty sync files provided, teardown process in synchronize is skipped and we simply update the objects


samirsilwal:
Remove result from sync-logs

Kushalkhadka7 (approved):

src/service/sync.ts
```diff
@@ -91,21 +118,36 @@ async function teardown(trx: Knex.Transaction, context: OperationContext): Promi
...
-    // Trigger onTeardownSuccess if bound.
-    if (context.params.onTeardownSuccess) {
-      await context.params.onTeardownSuccess({
-        connectionId,
-        data: null,
-        success: true,
-        timeElapsed: getElapsedTime(timeStart)
-      });
+    // Filter sync files to only include those that are available in the config and exclude any '.drop' files
+    const filteredSql = (syncFiles || []).filter(file => availableSql.includes(file) && !file.includes('.drop'));
```
mesaugat:
```suggestion
    const filesToSync = (syncFiles || []).filter(file => availableSql.includes(file) && !file.includes('.drop'));
```

src/util/string.ts
```diff
@@ -29,3 +29,40 @@ export function interpolate(template: string, params: any = {}): string {
...
+ * Converts CREATE statements to CREATE OR REPLACE/ALTER for supported object types.
+ * Supports: VIEW, FUNCTION, PROCEDURE
+ * - PostgreSQL/MySQL/Oracle: Uses CREATE OR REPLACE
+ * - MSSQL: Uses CREATE OR ALTER (SQL Server 2016+)
+ *
+ * @param {string} sql - The SQL content to modify
+ * @param {string} [dbClient] - Database client type ('pg', 'mssql', 'mysql', 'oracledb')
+ * @returns {string} - Modified SQL with CREATE OR REPLACE/ALTER syntax
+ */
+export function convertToCreateOrReplace(sql: string, dbClient?: string): string {
```
mesaugat: Will not need this as whoever is doing CREATE VIEW should change it on their own.

test/unit/service/sqlModifier.test.ts
```diff
@@ -0,0 +1,218 @@
+import 'mocha';
+import { expect } from 'chai';
+
+import * as sqlModifier from '../../../src/util/string';
+
+describe('SERVICE: sqlModifier', () => {
```
mesaugat: Don't need these tests as well.

src/service/sync.ts
```diff
@@ -2,30 +2,45 @@ import { Knex } from 'knex';
...
   const { connectionId } = context;
   const { hooks, sql } = context.config;
   const sqlBasePath = getSqlBasePath(context.config);
   const log = dbLogger(connectionId);
 
   log(`Running setup.`);
 
-  const sqlScripts = await sqlRunner.resolveFiles(sqlBasePath, sql);
+  if (isPartialSync && filteredSql.length === 0) {
+    log('No SQL files to synchronize in partial sync after filtering. Skipping setup.');
```
mesaugat:
```suggestion
    log('No SQL files to synchronize.');
```

test/unit/service/sync.test.ts
```diff
@@ -0,0 +1,94 @@
...
+      const syncFiles = ['function/schema/test_func.sql', 'function/schema/non_existent.sql'];
+      const availableSql = ['function/schema/test_func.sql', 'view/schema/test_view.sql'];
+
+      const filesToSync = syncFiles.filter(file => availableSql.includes(file) && !file.includes('.drop'));
+
+      expect(filesToSync).to.deep.equal(['function/schema/test_func.sql']);
+    });
+
+    it('should exclude .drop files from sync-files list', () => {
+      const syncFiles = ['function/schema/test_func.sql', 'function/schema/test_func.drop.sql'];
```
mesaugat:
```suggestion
      const syncFiles = ['function/schema/test_func.sql', 'function/schema/test_func.sql.drop'];
```

mesaugat (approved):

